PR #1, opened 2018-06-04
Fix URL of repository

znz:


kou:
@hsbt Can you permit me to write this repository?

kou:
@hsbt Thanks. Can you also set "Admin" permission like ruby/csv?

kou:
@hsbt Thanks. I've confirmed.

PR #6, opened 2018-09-26
Raise appropriate exception when failing to match end tag

zelivans:
In `BaseParser#pull_event`, when the ending tag can't be matched, `md` resolves to nil. The subsequent check of the value of `md[1]` will fail with `NoMethodError` (caught with `REXML::ParseException`). This proposed fix prevents the nil call and raises a `REXML::ParseException` directly.

```
/home/ariel/afl-kisaten/private/sandbox/rexml/rexml/lib/rexml/parsers/baseparser.rb:442:in `rescue in pull_event': #<NoMethodError: undefined method `[]' for nil:NilClass> (REXML::ParseException)
```

## Example input
With `REXML::Document.new` try:
```
<b></
```


kou:
Thanks.

PR #9, opened 2018-10-26
Rescue StandardError rather than Exception

byroot:
Ref: https://github.com/ruby/ruby/pull/1994

`Exception` is rescuing many things it shouldn't rescue like `SignalException` for instance.


kou:
Thanks!

PR #12, opened 2019-01-05
use #inspect to pretty-print meta-characters in error messages

gfx:
When input texts include non-printable characters, e.g. "\b" (`^H`),  REXML gives up parsing it, printing the non-printable characters as is. This will confuse users because the error messages in a terminal may be broken.

This PR fixes the problem by using `#inspect` to pretty-print those non-printable characters.


kou:
Good catch!

gfx:
Thanks 👍

PR #13, opened 2019-01-29
Fix crash with nil XPath variables

alyssais:
Prior to d91e124ab9843d2eaf62071bf029301254c36e2b, an XPath variable could be set to `nil`, in which case it would be silently converted to an empty string. This is clearly still the intended behaviour of this method, as can be seen from the (redundant) `when nil` further down. But, with the introduction of context support, this case was broken, because trying to do `@@context[:node]` would raise a `NoMethodError` if there was no context.

I also noticed that then `when nil` case was redundant, since without it it'll use `nil.to_s`, which is already the empty string, so I removed that in a seperate commit.

### DIFF
--- a/test/rexml/test_functions.rb
+++ b/test/rexml/test_functions.rb
@@ -6,6 +6,12 @@ require "rexml/document"
 module REXMLTests
   class FunctionsTester < Test::Unit::TestCase
     include REXML
+
+    def setup
+      super
+      REXML::Functions.context = nil
+    end
+
     def test_functions
       # trivial text() test
       # confuse-a-function
@@ -223,9 +229,6 @@ module REXMLTests
     end
 
     def test_string_nil_without_context
-      # Reset to default value
-      REXML::Functions.class_variable_set(:@@context, nil)
-
       doc = REXML::Document.new(<<-XML)
       <?xml version="1.0" encoding="UTF-8"?>
       <root>
--- a/test/rexml/test_functions.rb
+++ b/test/rexml/test_functions.rb
@@ -222,6 +222,22 @@ module REXMLTests
       assert_equal( [REXML::Comment.new("COMMENT A")], m )
     end
 
+    def test_string_nil_without_context
+      # Reset to default value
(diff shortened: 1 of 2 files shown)


kou:
Thanks.
Could you add a test for this case?

alyssais:
Actually, is there any reason `@@context` shouldn't be initialized to `{}`? That would probably be a more robust fix for this issue.

That wouldn't really be able to be tested, though, because there's no behaviour. Just a change in default value, which a single unit test in the middle of a run couldn't check. What do you think?

If not, I've added a test for the current fix.

test/rexml/test_functions.rb
```diff
@@ -222,6 +222,22 @@ def test_normalize_space
       assert_equal( [REXML::Comment.new("COMMENT A")], m )
     end
 
+    def test_string_nil_without_context
+      # Reset to default value
+      REXML::Functions.class_variable_set(:@@context, nil)
```
kou: We can use `REXML::Functions.context = nil`.

Could you create a setup method and call this? Because this is needed for all tests.

alyssais:
> -    def Functions::string( object=nil )
> -      object = @@context[:node] if object.nil?
> +    def Functions::string( object=NOT_SPECIFIED )
> +      object = @@context[:node] if NOT_SPECIFIED == object

Could this just be the following?:

    def Functions::string( object=@@context[:node] )

test/rexml/test_functions.rb
alyssais: All tests in this file? Or the entire test suite?

kou:
It will work too.

test/rexml/test_functions.rb
kou: It's enough only for this file.

gsar:
@kou any plans to merge this fix into the next ruby release? thanks!

kou:
This will be included in Ruby 2.7.0.
If you can't wait Ruby 2.7.0, you can use rexml 3.2.1 gem.

PR #16, opened 2019-02-19
Message less confusing error to human

ujihisa:
Message less confusing error to human

* Problem: Following error message is not helpful, because you have to reason
  that '' actually means it's in the top-level, and the 'div' (not '</div>') is
  an end tag

        require "rexml/parsers/lightparser"
        REXML::Parsers::LightParser.new('</div>').parse
        #=> Missing end tag for '' (got 'div')

* Solution: add a special case in error handling just to change the error message

        require "rexml/parsers/lightparser"
        REXML::Parsers::LightParser.new('</div>').parse
        #=> Unexpected top-level end tag (got 'div')

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -336,8 +336,7 @@ module REXML
               last_tag = @tags.pop
               md = @source.match( CLOSE_MATCH, true )
               if md and !last_tag
-                message = "Unexpected top-level end tag"
-                message << " (got '#{md[1]}')" if md
+                message = "Unexpected top-level end tag (got '#{md[1]}')"
                 raise REXML::ParseException.new(message, @source)
               end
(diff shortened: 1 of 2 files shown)


lib/rexml/parsers/baseparser.rb
```diff
@@ -335,6 +335,10 @@ def pull_event
               @nsstack.shift
               last_tag = @tags.pop
               md = @source.match( CLOSE_MATCH, true )
+              if md && !last_tag
```
kou: Could you use `and`?

ujihisa: Definitely. Sorry I missed using `and` which is commonly used in this project.
I rebase'ed to use `and` instead of `&&` at my first commit.

kou:
Thanks.
I've merged.

ujihisa:
thx!

PR #23, opened 2020-01-15
Add GitHub Actions script

slonopotamus:
Unlike current Travis setup, this tests much more operating systems.

You may see that [it currently doesn't pass on JRuby](https://github.com/slonopotamus/rexml/commit/4cf0be5e7c3fab00392f15bf26e360ac9550f1a7/checks?check_suite_id=402610859). Unfortunately, fixing that is beyond my current Ruby skills. I see two options here - either someone helps with fixing tests on JRuby or I exclude JRuby from current PR and file a separate issue for fixing it.

Note: Ruby 2.3 is omitted, but given that it [reached EOL](https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/) almost a year ago, it doesn't look like a big loss.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,28 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  build:
+    strategy:
+      matrix:
+        ruby: [2.4, 2.5, 2.6, 2.7, jruby]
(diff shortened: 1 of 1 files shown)


kou:
Sorry for not responding this.
I've done.

PR #24, opened 2020-01-16
Add actions, two test fixes for newer Rubies

MSP-Greg:
Actions - Ubuntu, macOS, Windows, Ruby 2.3 thru 2.7, except 2.3 on Windows

rexml is freezing in ruby/ruby macOS CI.  Wanted to check here on 2.7, etc.

Minimal changes to two test/lib files, just enough to stop extraneous output.


kou:
Sorry for not responding this.
I've done.

PR #37, opened 2020-12-09
Refine dclone definitions not to leak outside of REXML

amatsuda:
While investigating my app, I found an unfamiliar `dclone` method on `Object`, which turned out to be defined here in `xpath_parser.rb` only for using in `xpath_parser.rb`.

I believe this is the very use case of Ruby's most underevaluated feature, Refinements.

With this patch, we're no longer seeing the `dclone` method leaking into our apps.


kou:
Thanks!

PR #39, opened 2021-01-06
Exclude useless files after installation

nobu:
These files are meaningless outside the working directory.

### DIFF
--- a/rexml.gemspec
+++ b/rexml.gemspec
@@ -17,7 +17,6 @@ Gem::Specification.new do |spec|
   spec.license       = "BSD-2-Clause"
 
   spec.files         = [
-    "Gemfile",
     "LICENSE.txt",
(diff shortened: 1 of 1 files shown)


rexml.gemspec
```diff
@@ -17,8 +17,6 @@ Gem::Specification.new do |spec|
   spec.license       = "BSD-2-Clause"
 
   spec.files         = [
-    ".gitignore",
-    ".travis.yml",
     "Gemfile",
```
kou, line 20: Could you also remove this and `Rakefile`?

nobu: Indeed, it is useless too.

kou:
```suggestion
```

kou:
Thanks.

PR #44, opened 2021-02-04
Context

BurdetteLamar:
This new page, "Element Context," will provide a good link target for all other mentions of context.

The discussion of <tt>:raw</tt> is inadequate, but still maybe better than nothing.

### DIFF
--- a/doc/rexml/context.rdoc
+++ b/doc/rexml/context.rdoc
@@ -1,4 +1,4 @@
-== Context
+== Element Context
 
 Notes:
 - All code on this page presupposes that the following has been executed:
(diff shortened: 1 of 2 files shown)


doc/rexml/context.rdoc
```diff
@@ -0,0 +1,140 @@
+== Element Context
+
+Notes:
+- All code on this page presupposes that the following has been executed:
+
+    require 'rexml/document'
+    include REXML
```
kou, line 7: I think that this may confuse readers.
How about using `REXML::` explicitly like `REXML::Document`?

BurdetteLamar: Done.

kou:
Thanks!

PR #45, opened 2021-02-05
Enhanced RDoc for Element

BurdetteLamar:
A little prefatory material and some tweaks to previously revised methods doc.


kou:
Thanks!

PR #46, opened 2021-02-05
Enhanced RDoc for Element

BurdetteLamar:
Revises methods
- whitespace
- ignore_whitespace_nodes
- raw

Please consider whether these may be too spare.


kou:
Thanks!

I think that referring existing document for details makes sense.

PR #47, opened 2021-02-06
Enhanced RDoc for Document

BurdetteLamar:
Revises prefatory material, and methods:
- new
- node_type
- clone
- expanded_name
- add

Note well:  When I run this through rdoc, the call-seq for method expanded_name, and its aliased method name, does not show up.  I have no theory that explains that.

### DIFF
--- a/lib/rexml/document.rb
+++ b/lib/rexml/document.rb
@@ -112,9 +112,6 @@ module REXML
       Document.new self
     end
 
-    # :call-seq:
-    #   expanded_name -> empty_string
-    #
     # Returns an empty string.
     #
     def expanded_name
--- a/lib/rexml/document.rb
+++ b/lib/rexml/document.rb
@@ -14,25 +14,72 @@ require_relative "parsers/streamparser"
 require_relative "parsers/treeparser"
 
 module REXML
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
@kou, did you see this one?  I left it as a draft b/c it has a problem with a call-seq.

kou:
It seems that the `expanded_name` and `name` problem is a bug of RDoc. Could you report it to RDoc?
It seems that the `name` method name may be related.

BurdetteLamar:
A workaround for the rdoc bug is to omit the call-seq for expanded_name, which is now done here.

I'll have to study how to correctly submit a bug for RDoc.

BurdetteLamar:
@kou, now ready for review.

BurdetteLamar:
@kou said:

> 
> 
> It seems that the `expanded_name` and `name` problem is a bug of RDoc. Could you report it to RDoc?
> It seems that the `name` method name may be related.

https://github.com/ruby/rdoc/issues/792

BurdetteLamar:
Restored the call-seq for Document#expanded_name.

kou:
Thanks!

PR #49, opened 2021-02-07
Enhanced RDoc for Document

BurdetteLamar:
Revises methods:
- add_element
- root
- doctype
- xml_decl
- version
- encoding
- stand_alone?


kou:
Thanks!

PR #50, opened 2021-02-08
Enhanced RDoc for Elements

BurdetteLamar:
Revises methods:
- new
- []
- []=
- empty?
- index
- delete
- delete_all


kou:
Thanks!

PR #57, opened 2021-02-14
Enhanced RDoc for Element

BurdetteLamar:
Revises methods:
- has_text?
- text
- get_text
- text=
- add_text
- node_type
- xpath


kou:
Thanks!

Issue #58, opened 2021-02-15
Parent#index never returns nil


BurdetteLamar:
Should return nil when:
- No children.
- Child not found.

I have a fix for this in a forthcoming PR.

BurdetteLamar:
Reviewing this.

PR #59, opened 2021-02-15
Fix for issue #58, Parent#index never returns nil

BurdetteLamar:
I'm putting the test into a new file, test_parent.rb.  I think there are more bugs in Parent, and so there will be more tests.


BurdetteLamar:
Reviewing this.

PR #60, opened 2021-02-16
Code fix, test, and RDoc for Parent#index

BurdetteLamar:
The method failed to return nil both for an empty parent, and for parent not containing the child.

I've created file test/test_parent.rb because there will be more tests for this class, and I think there are a couple more bugs (think insert_before and insert_after).

### DIFF
--- a/lib/rexml/parent.rb
+++ b/lib/rexml/parent.rb
@@ -116,14 +116,23 @@ module REXML
       @children.dup
     end
 
-    # Fetches the index of a given child
-    # @param child the child to get the index of
-    # @return the index of the child, or nil if the object is not a child
-    # of this parent.
-    def index( child )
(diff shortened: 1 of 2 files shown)


BurdetteLamar:
@kou, are you seeing this?

kou:
I took a look at this and found some points that I need to left comments.
But I don't have enough time for it for a few days. Please wait for a while.

BurdetteLamar:
> 
> 
> I took a look at this and found some points that I need to left comments.
> But I don't have enough time for it for a few days. Please wait for a while.

Thanks, @kou.  I will find other work.

BurdetteLamar:
I will re-submit this next week.  (My PR pipeline is fouled.)

PR #61, opened 2021-02-19
Enhanced RDoc for Element

BurdetteLamar:
Marking this as a draft b/c it has a couple of TODOs that need review and resolution.

Revised methods:
- []
- attribute
- has_attributes?
- add_attribute
- add_attributes
- delete_attribute
- cdatas
- comments
- instructions
- texts

### DIFF
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -945,19 +945,45 @@ module REXML
     # Attributes                                    #
     #################################################
 
-    # Fetches an attribute value or a child.
+    # :call-seq:
+    #   [index] -> object
+    #   [attr_name] -> attr_value
+    #   [attr_sym] -> attr_value
     #
-    # If String or Symbol is specified, it's treated as attribute
-    # name. Attribute value as String or +nil+ is returned. This case
-    # is shortcut of +attributes[name]+.
+    # With integer argument +index+ given,
+    # returns the child at offset +index+, or +nil+ if none:
     #
-    # If Integer is specified, it's treated as the index of
-    # child. It returns Nth child.
+    #   d = REXML::Document.new '><root><a/>text<b/>more<c/></root>'
+    #   root = d.root
+    #   (0..root.size).each do |index|
+    #     node = root[index]
+    #     p "#{index}: #{node} (#{node.class})"
+    #   end
     #
-    #   doc = REXML::Document.new("<a attr='1'><b/><c/></a>")
-    #   doc.root["attr"]             # => "1"
-    #   doc.root.attributes["attr"]  # => "1"
-    #   doc.root[1]                  # => <c/>
+    # Output:
+    #
+    #   "0: <a/> (REXML::Element)"
+    #   "1: text (REXML::Text)"
+    #   "2: <b/> (REXML::Element)"
+    #   "3: more (REXML::Text)"
+    #   "4: <c/> (REXML::Element)"
+    #   "5:  (NilClass)"
+    #
+    # With string argument +attr_name+ given,
+    # returns the string value for the given attribute name if it exists,
+    # otherwise +nil+:
+    #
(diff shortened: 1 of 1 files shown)


lib/rexml/element.rb
```diff
@@ -1035,19 +1121,19 @@ def add_attributes hash
...
+    #
+    # Removes a named attribute if it exists;
+    # returns the removed attribute if found, otherwise +nil+:
+    #
+    #   e = REXML::Element.new('foo')
+    #   e.add_attribute('bar', 'baz')
+    #   e.delete_attribute('bar') # => <bar/>
+    #   e.delete_attribute('bar') # => nil
+    #
+    # TODO: Determine whether delete_attribute(attribute) actually works.
```
kou, line 1135: What do we need to do to resolve this?

lib/rexml/element.rb
```diff
@@ -969,6 +995,30 @@ def [](name_or_index)
...
+    #       <c attr="value"/>
+    #    </root>
+    #   EOT
+    #   d = REXML::Document.new(xml_string)
+    #   root = d.root
+    #   a = root[1] # => <a xmlns='ns1' attr='value'/>
+    #   a.attribute('attr') # => attr='value'
+    #   a.attribute('nope') # => nil
+    #
+    # TODO:  Make sensible examples using both arguments.
```
kou, line 1020: How about this?

```text
require "rexml/document"

xml = <<-XML
<root xmlns:a='a' a:x='a:x' x='x'/>
XML
document = REXML::Document.new(xml)
document.root.attribute("x")      # => x='x'
document.root.attribute("x", "a") # => a:x='a:x'
```

lib/rexml/element.rb
BurdetteLamar: Did not work for me (didn't delete the given REXML::Attribute object).  Resolution is either an example that works or an agreement to omit because it doesn't work.  My personal opinion is that this was not meant to work and is a documentation error.

kou: Ah, the example doesn't work as expected?
Could you clarify the expected and the actual? For example, is the example expected or actual?

BurdetteLamar: Sorry for this poorly framed issue.  The old doc says that the argument can be "either an Attribute or a String."  An Attribute argument did not work for me:

      e = REXML::Element.new('foo') # => <foo/>
      e.attributes # => {}
      a = REXML::Attribute.new('bar', 'baz') # => bar='baz'
      e.add_attribute(a) # => bar='baz'
      e.attributes # => {"bar"=>bar='baz'}
      e.delete_attribute(a) # => nil
      e.attributes # => {"bar"=>bar='baz'}

lib/rexml/element.rb
BurdetteLamar: Added.  Please see whether you think the accompanying text is adequate.

kou:
Please wait for a few days...

BurdetteLamar:
Happy  to wait, thanks.

BurdetteLamar:
@kou , do you expect to review this week?

lib/rexml/element.rb
kou: Confirmed. `e.delete_attribute(a)` doesn't work with the oldest REXML that I have.
We don't support it.

kou (commented):
Sorry for my late response...

lib/rexml/element.rb
kou: Confirmed. It looks good to me.

BurdetteLamar:
> 
> 
> Sorry for my late response...

NP, my friend.  We're all busy.

kou:
Thanks!

BurdetteLamar:
> 
> 
> Thanks!

Is there someplace where the doc for rexml/master is built out?

kou:
https://ruby.github.io/rexml/

PR #62, opened 2021-03-01
Enhanced RDoc for Element

BurdetteLamar:
This is more robust replacement for the currently brief doc for class Element.

### DIFF
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -21,8 +21,10 @@ module REXML
   #
   # - Has a name (string).
   # - May have a parent (another element).
-  # - Has zero or more children (other elements, text, CDATA, and comments).
-  # - Has zero or more siblings (other elements, text, CDATA, and comments).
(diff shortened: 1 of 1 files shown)


lib/rexml/element.rb
```diff
@@ -15,9 +15,256 @@ module REXML
...
-  # having children, attributes, and names, and can themselves be
-  # children.
+  # An \REXML::Element object represents an XML element.
+  #
+  # An element:
+  #
+  # - Has a name (string).
+  # - May have a parent (another element).
+  # - Has zero or more children (other elements, text, CDATA, and comments).
+  # - Has zero or more siblings (other elements, text, CDATA, and comments).
```
kou, line 25: Could you also add processing instructions to children and siblings?

BurdetteLamar: Done.

kou:
Thanks!

PR #64, opened 2021-03-07
Task-oriented documentation

BurdetteLamar:
The important page is the master table of contents in ..\doc\rexml\tasks\tocs\master_toc.rdoc
The idea is to let the user find the relevant task, then follow a link to see code that does that.

I started out calling these _recipes_, but so far each is little, so I think _task_ is a better term.

The best way to view these is to run RDoc and begin at the master toc html page.


doc/rexml/tasks/mktocs.rb
```diff
@@ -0,0 +1,72 @@
+Dir.chdir(File.dirname(__FILE__)) do
```
kou, line 1: Could you define this as a `doc:toc:generate` Rake task in `Rakefile`?

BurdetteLamar: > 
> 
> Could you define this as a `rdoc:toc:generate` Rake task in `Rakefile`?

Sorry, I meant to say more (and better).  My initial question is Do we want this sort of thing?

If yes, everything else can come later:  correctness, robustness of TOC build, etc.

Also, Element is not completed, so there are some hanging issues there.

kou: It's acceptable.

It's better that RDoc provides the TOC feature as a build-in feature. For example, `[:toc:]` is expanded to TOC on rendering phase. But it seems that RDoc doesn't provide the TOC feature. (Could you request the feature to RDoc?)

So it's reasonable that we have TOC generator for us for now.

BurdetteLamar: > 
> 
> It's acceptable.
> 
This is a *lot* of work, so needs to be much more than 'acceptable.' Do you like the 'task' concept?

kou: Yes. I like the concept.

BurdetteLamar: Rakefile updated with task  tocs:generate.

kou:
Can we merge this pull request?
Do you still want to work on this pull request?

BurdetteLamar:
> 
> 
> Can we merge this pull request?
> Do you still want to work on this pull request?

Fine with me to merge.  I'll continue in a new PR.

kou:
OK. I've merged.

doc/rexml/tasks/mktocs.rb
BurdetteLamar: > 
> 
> It's acceptable.
> 
> It's better that RDoc provides the TOC feature as a build-in feature. For example, `[:toc:]` is expanded to TOC on rendering phase. But it seems that RDoc doesn't provide the TOC feature. (Could you request the feature to RDoc?)
> 
> So it's reasonable that we have TOC generator for us for now.

@kou, I have created the feature request: https://github.com/ruby/rdoc/issues/801

kou: Thanks!

PR #82, opened 2021-12-17
Access constants with :: instead of . in DTD::Parser.

camertron:
I was trying to use REXML's DTD parser today. Here's what I did:

```ruby
require 'rexml/dtd/dtd'
dtd = REXML::DTD::Parser.parse(File.read('/path/to/file.dtd'))
```

The code above produced the following error:

/Users/camertron/.asdf/installs/ruby/3.0.2/lib/ruby/gems/3.0.0/gems/rexml-3.2.5/lib/rexml/dtd/dtd.rb:26:in \`parse_helper': undefined method \`PATTERN_RE' for REXML::DTD::ElementDecl:Class (NoMethodError)

I took a peek at the code and discovered that `PATTERN_RE` is indeed not a method but instead a constant, and should be looked up using the `::` syntax instead of `.`.

It seems a little strange that nobody's noticed this before, so... am I missing something?

### DIFF
--- a/lib/rexml/dtd/dtd.rb
+++ b/lib/rexml/dtd/dtd.rb
@@ -23,19 +23,19 @@ module REXML
         contents = Parent.new
         while input.size > 0
           case input
-          when ElementDecl.PATTERN_RE
+          when ElementDecl::PATTERN_RE
             match = $&
             contents << ElementDecl.new( match )
-          when AttlistDecl.PATTERN_RE
+          when AttlistDecl::PATTERN_RE
             matchdata = $~
             contents << AttlistDecl.new( matchdata )
-          when EntityDecl.PATTERN_RE
+          when EntityDecl::PATTERN_RE
             matchdata = $~
             contents << EntityDecl.new( matchdata )
-          when Comment.PATTERN_RE
+          when Comment::PATTERN_RE
             matchdata = $~
             contents << Comment.new( matchdata )
-          when NotationDecl.PATTERN_RE
+          when NotationDecl::PATTERN_RE
             matchdata = $~
             contents << NotationDecl.new( matchdata )
           end


olleolleolle:
I took a quick look in the tests directory and was not able to find a test that ran this code. Can you find or add one?

kou:
This code isn't used.

In normal case (XML parsing case), DTD is parsed in `REXML::Parsers::BaseParser#pull_event`.

Do you have any use case that you want to parse only DTD?

camertron:
@olleolleolle nice, that looks like a great start :) Unfortunately I don't really have the bandwidth to contribute additional tests right now. This issue is just barely a step above a syntax error IMHO though, so hopefully tests aren't a blocker?

@kou my use-case is parsing the [DTD for the Locale Data Markup Language (LDML)](https://github.com/unicode-org/cldr/blob/main/common/dtd/ldml.dtd) which comes with the Common Locale Data Repository (CLDR) data set. I need to be able to iterate over the various possible attributes for certain elements in locale-specific XML files like [this one for English](https://github.com/unicode-org/cldr/blob/main/common/main/en.xml). I haven't been able to figure out how to parse DTDs with Nokogiri, so I thought I'd give rexml a try and ran into this `NoMethodError`.

It's worth mentioning that, even with the method call converted into a constant lookup, rexml appears to hang when parsing the LDML DTD, even when I give it only small bits. I'm not sure why. It sounds like there is other code in rexml that inspects the DTD when parsing an XML document, is it possible to use that code independently of the XML parser?

kou:
I don't recommend that you use `REXML::DTD::Parser` because it's not maintained.

How about parsing the DTD inside an XML?

```ruby
require "rexml/document"

dtd = File.read("ldml.dtd")
xml = <<-XML
<!DOCTYPE ldml [
#{dtd}
]>
<ldml/>
XML
document = REXML::Document.new(xml)
pp document.doctype
```

camertron:
Hey @kou, that works very well, thanks!

PR #87, opened 2022-06-06
Fix RDoc for Element

aleksandrilyin:
* Add missing plus for `Element#has_text?`.
* Remove unneeded hash and duplicated `the` for `Element#text`.

### DIFF
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -989,7 +989,7 @@ module REXML
     # :call-seq:
     #   has_text? -> true or false
     #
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
👍 Good finds! Thanks for the PR!

olleolleolle:
The build failure on Windows, #88 I reported here, to keep it visible. The cause is probably elsewhere.

aleksandrilyin:
> The build failure on Windows, #88 I reported here, to keep it visible. The cause is probably elsewhere.

Thank you.

PR #93, opened 2023-01-20
CI against Ruby 3.0, 3.1, and 3.2

amatsuda:
Here's a patch that adds CI configurations for Rubies 3.0, 3.1, and 3.2 that are currently untested.


kou:
Thanks!

PR #96, opened 2023-05-25
xpath abbreviate: add support for string literal that contains double-quote

pulver:
This adds support for a string literal that contains a double-quote to `XPathParser#abbreviate`. Basically any literal that contains a double-quote `"` must be quoted by single-quotes `'` since XPath 1.0 does not support any escape characters.

The change improves the following test script
```ruby
require 'rexml'

parsed = REXML::Parsers::XPathParser.new.parse('/a[b/text()=concat("c\'",\'"d\')]')
puts "#{parsed}"
puts ""
appreviated = REXML::Parsers::XPathParser.new.abbreviate parsed
puts "#{appreviated}"
```
### Output Before Change
```
[:document, :child, :qname, "", "a", :predicate, [:eq, [:child, :qname, "", "b", :child, :text], [:function, "concat", [[:literal, "c'"], [:literal, "\"d"]]]]]

/a[ b/text() = concat( "c'" , "\"d" ) ]
```
### Output After Change
```
[:document, :child, :qname, "", "a", :predicate, [:eq, [:child, :qname, "", "b", :child, :text], [:function, "concat", [[:literal, "c'"], [:literal, "\"d"]]]]]

/a[ b/text() = concat( "c'" , '"d' ) ]
```

### DIFF
--- a/lib/rexml/parsers/xpathparser.rb
+++ b/lib/rexml/parsers/xpathparser.rb
@@ -188,6 +188,7 @@ module REXML
         return string.squeeze(" ")
       end
 
+      private
       def quote_literal( literal )
         case literal
         when String
@@ -203,7 +204,6 @@ module REXML
         end
       end
 
-      private
       #LocationPath
       #  | RelativeLocationPath
       #  | '/' RelativeLocationPath?
--- a/lib/rexml/parsers/xpathparser.rb
+++ b/lib/rexml/parsers/xpathparser.rb
@@ -178,7 +178,7 @@ module REXML
         when :literal
           path.shift
           string << " "
-          string << path.shift.inspect
+          string << QuoteLiteral(path.shift)
           string << " "
         else
           string << " "
@@ -361,6 +361,18 @@ module REXML
           OrExpr(pred, preds)
         }
         path
+      end
+
+      def QuoteLiteral literal
+        case literal
+        when String
+          # Xpath 1.0 does not support escape characters.
+          # Assumes literal does not contain both single and double quotes.
+          pattern = literal.include?('"') ? "'%s'" : '"%s"'
+          pattern % literal
+        else
+          literal.inspect
+        end
       end
 
       # The following return arrays of true/false, a 1-1 mapping of the
--- a/lib/rexml/parsers/xpathparser.rb
+++ b/lib/rexml/parsers/xpathparser.rb
@@ -178,7 +178,7 @@ module REXML
         when :literal
           path.shift
           string << " "
-          string << QuoteLiteral(path.shift)
+          string << quote_literal(path.shift)
           string << " "
         else
           string << " "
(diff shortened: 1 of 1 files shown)


pulver:
Good catch, thank you. I restored the previous method of `literal.inspect` for all non-`String`s.

I wasn't sure how to handle this `literal` translation so I left it as it was:

https://github.com/ruby/rexml/blob/54b7109172bbe36a6702b3844913d715d65ebe9c/lib/rexml/parsers/xpathparser.rb#L86-L87

lib/rexml/parsers/xpathparser.rb
```diff
@@ -363,6 +363,18 @@ def Predicate path, parsed
         path
       end
 
+      def QuoteLiteral literal
```
kou, line 366: Could you use `snake_case`?

```suggestion
      def quote_literal(literal)
```

lib/rexml/parsers/xpathparser.rb
```diff
@@ -363,6 +363,18 @@ def Predicate path, parsed
...
+        case literal
+        when String
+          # Xpath 1.0 does not support escape characters.
```
kou, line 369:
```suggestion
          # XPath 1.0 does not support escape characters.
```

lib/rexml/parsers/xpathparser.rb
```diff
@@ -363,6 +363,18 @@ def Predicate path, parsed
...
+          # Assumes literal does not contain both single and double quotes.
+          pattern = literal.include?('"') ? "'%s'" : '"%s"'
+          pattern % literal
```
kou, line 372: Could you use string interpolation instead of `String#%`?

```suggestion
          if literal.include?("'")
            "\"#{literal}\""
          else
            "'#{literal}'"
          end
```

lib/rexml/parsers/xpathparser.rb
kou: And could you move this method to before `#LocalPath` (the line 192)?
Method in `#LocationPath` to `#FunctionCall` are for syntaxes in XPath. I don't want to mix this and them.

pulver:
Done, thanks for the feedback.

lib/rexml/parsers/xpathparser.rb
```diff
@@ -188,6 +188,21 @@ def predicate_to_string( path, &block )
...
+            "\"#{literal}\""
+          else
+            "'#{literal}'"
+          end
+        else
+          literal.inspect
+        end
+      end
+
       private
```
kou, line 206: Could you move this `private` above `quote_literal`?

pulver: Done: d6a02b91a244b8c0805a780525568ad4bc2b74de

kou (approved):

PR #103, opened 2024-01-05
Use reusing workflow for Ruby versions

hsbt:
This automatically add new version of Ruby for GitHub Actiosn.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,14 @@ on:
   - push
   - pull_request
 jobs:
+  ruby-versions:
+    uses: ruby/actions/.github/workflows/ruby_versions.yml@master
+    with:
+      engine: cruby-jruby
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -13,15 +20,7 @@ jobs:
...
-        ruby-version:
-          - "2.5"
-          - "2.6"
-          - "2.7"
-          - "3.0"
-          - "3.1"
-          - "3.2"
-          - "3.3"
-          - jruby
+          ruby-version: ${{ fromJson(needs.ruby-versions.outputs.versions) }}
```
kou, line 23:
```suggestion
        ruby-version: ${{ fromJson(needs.ruby-versions.outputs.versions) }}
```

hsbt: Oh, thanks!

PR #104, opened 2024-01-06
Add parse benchmark

naitoh:
I want to improve the parsing process and would like to add a parsing benchmark.

The benchmark process just parses the XML from beginning to end.

Since performance differs depending on whether YJIT is ON or OFF, both are measured.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -28,3 +28,42 @@ RDoc::Task.new do |rdoc|
 end
 
 load "#{__dir__}/tasks/tocs.rake"
+
+benchmark_tasks = []
+namespace :benchmark do
+  Dir.glob("benchmark/*.yaml").sort.each do |yaml|
+    name = File.basename(yaml, ".*")
+    env = {
+      "RUBYLIB" => nil,
+      "BUNDLER_ORIG_RUBYLIB" => nil,
+    }
+    command_line = [
+      RbConfig.ruby, "-v", "-S", "benchmark-driver", File.expand_path(yaml),
+    ]
+
+    desc "Run #{name} benchmark"
+    task name do
+      puts("```")
+      sh(env, *command_line)
+      puts("```")
+    end
+    benchmark_tasks << "benchmark:#{name}"
+
+    case name
+    when /\Aparse/
+      namespace name do
+        desc "Run #{name} benchmark: small"
+        task :small do
+          puts("```")
+          sh(env.merge("N_ELEMENTS" => "500", "N_ATTRIBUTES" => "1"),
+             *command_line)
+          puts("```")
+        end
+        benchmark_tasks << "benchmark:#{name}:small"
+      end
+    end
+  end
+end
+
+desc "Run all benchmarks"
+task :benchmark => benchmark_tasks
--- a/benchmark/parse.yaml
+++ b/benchmark/parse.yaml
@@ -1,0 +1,57 @@
+loop_count: 100
+contexts:
+  - gems:
+      rexml: 3.2.6
+    require: false
+    prelude: require 'rexml'
+  - name: master
+    prelude: |
+      $LOAD_PATH.unshift(File.expand_path("lib"))
+      require 'rexml'
+  - name: 3.2.6(YJIT)
+    gems:
+      rexml: 3.2.6
+    require: false
+    prelude: |
+      require 'rexml'
+      RubyVM::YJIT.enable
+  - name: master(YJIT)
+    prelude: |
+      $LOAD_PATH.unshift(File.expand_path("lib"))
+      require 'rexml'
+      RubyVM::YJIT.enable
+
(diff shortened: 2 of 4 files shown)


rexml.gemspec
```diff
@@ -57,5 +57,6 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "bundler"
   spec.add_development_dependency "rake"
+  spec.add_development_dependency "benchmark_driver"
   spec.add_development_dependency "test-unit"
```
kou, line 61: Could you keep this list in alphabetical order?

benchmark/parse.yaml
```diff
@@ -0,0 +1,53 @@
...
+      RubyVM::YJIT.enable
+
+prelude: |
+  require 'rexml/document'
+  require 'rexml/parsers/sax2parser'
+  require 'rexml/parsers/pullparser'
+  require 'rexml/parsers/streamparser'
+  require 'rexml/streamlistener'
+
+  n_lines = Integer(ENV.fetch("N_LINES", "10000"), 10)
```
kou, line 31: `n_elements`?

benchmark/parse.yaml
```diff
@@ -0,0 +1,53 @@
...
+  xml = build_xml(n_lines, n_attributes)
+
+  class Listener
+    include REXML::StreamListener
+  end
+
+benchmark:
+  'dom'        : REXML::Document.new(xml).elements.each("root/child") {|_|}
+  'sax'        : REXML::Parsers::SAX2Parser.new(xml).parse
+  'pull'       : p = REXML::Parsers::PullParser.new(xml); while p.has_next?; res = p.pull; end
```
kou, line 52: Could you use multiple lines with `pull: |` or something instead of one liner for easy to read?

naitoh: I see. I'll fix it!

benchmark/parse.yaml
naitoh: Yes. I'll fix it!
Thanks!

rexml.gemspec
naitoh: OK. I'll fix it!
Thanks!

Rakefile
```diff
@@ -28,3 +28,42 @@ RDoc::Task.new do |rdoc|
...
+    desc "Run #{name} benchmark"
+    task name do
+      puts("```")
+      sh(env, *command_line)
+      puts("```")
+    end
+    benchmark_tasks << "benchmark:#{name}"
+
+    case name
+    when /\Aparse/, "shift"
```
kou, line 53:
```suggestion
    when /\Aparse/
```

benchmark/parse.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+  class Listener
+    include REXML::StreamListener
+  end
+
+benchmark:
+  'dom'        : REXML::Document.new(xml).elements.each("root/child") {|_|}
+  'sax'        : REXML::Parsers::SAX2Parser.new(xml).parse
+  'pull'       : |
+    p = REXML::Parsers::PullParser.new(xml)
```
kou, line 53: How about using `parser` for variable name because we have `p` method?

```suggestion
    parser = REXML::Parsers::PullParser.new(xml)
```

benchmark/parse.yaml
```diff
@@ -0,0 +1,57 @@
...
+    while p.has_next?
+      res = p.pull
```
kou, line 55: It seems that we don't need to assign the return value:

```suggestion
      p.pull
```

benchmark/parse.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+prelude: |
+  require 'rexml/document'
+  require 'rexml/parsers/sax2parser'
+  require 'rexml/parsers/pullparser'
+  require 'rexml/parsers/streamparser'
+  require 'rexml/streamlistener'
+
+  n_elements = Integer(ENV.fetch("N_ELEMENTS", "10000"), 10)
+  n_attributes = Integer(ENV.fetch("N_ATTRIBUTES", "10"), 10)
```
kou, line 32: With these configurations, we take 10+ minutes for one `rake benchmark`: https://github.com/ruby/rexml/actions/runs/7431570516/job/20222471649?pr=104

Can we reduce benchmark time? Or do we need these numbers for meaningful benchmark?

naitoh: I set this value because I need to execute XML with a large number of elements, but the execution time is too long, so I will reduce the number of elements.

naitoh: https://github.com/ruby/rexml/actions/runs/7434389228/job/20228488163?pr=104
Reduced to 1/4.

kou:
Thanks!

PR #114, opened 2024-02-17
Use more StringScanner based API to parse XML

naitoh:
## Why?

Improve maintainability by optimizing the process so that the parsing process proceeds using StringScanner#scan.

## Changed
- Change `REXML::Parsers::BaseParser` from `frozen_string_literal: false` to `frozen_string_literal: true`.
- Added `Source#string=` method for error message output.
- Added TestParseDocumentTypeDeclaration#test_no_name test case.
- Of the `intSubset` of DOCTYPE, "<!" added consideration for processing `Comments` that begin with "<!".

## [Benchmark]

```
RUBYLIB= BUNDLER_ORIG_RUBYLIB= /Users/naitoh/.rbenv/versions/3.3.0/bin/ruby -v -S benchmark-driver /Users/naitoh/ghq/github.com/naitoh/rexml/benchmark/parse.yaml
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [arm64-darwin22]
Calculating -------------------------------------
                         before       after  before(YJIT)  after(YJIT)
                 dom     11.240      10.569        17.173       18.219 i/s -     100.000 times in 8.896882s 9.461267s 5.823007s 5.488884s
                 sax     31.812      30.716        48.383       52.532 i/s -     100.000 times in 3.143500s 3.255655s 2.066861s 1.903600s
                pull     36.855      36.354        56.718       61.443 i/s -     100.000 times in 2.713300s 2.750693s 1.763099s 1.627523s
              stream     34.176      34.758        49.801       54.622 i/s -     100.000 times in 2.925991s 2.877065s 2.008003s 1.830779s

Comparison:
                              dom
         after(YJIT):        18.2 i/s
        before(YJIT):        17.2 i/s - 1.06x  slower
              before:        11.2 i/s - 1.62x  slower
               after:        10.6 i/s - 1.72x  slower

                              sax
         after(YJIT):        52.5 i/s
        before(YJIT):        48.4 i/s - 1.09x  slower
              before:        31.8 i/s - 1.65x  slower
               after:        30.7 i/s - 1.71x  slower

                             pull
         after(YJIT):        61.4 i/s
        before(YJIT):        56.7 i/s - 1.08x  slower
              before:        36.9 i/s - 1.67x  slower
               after:        36.4 i/s - 1.69x  slower

                           stream
         after(YJIT):        54.6 i/s
        before(YJIT):        49.8 i/s - 1.10x  slower
               after:        34.8 i/s - 1.57x  slower
[...]

PR #117, opened 2024-03-02
Remove `Source#string=` method

naitoh:
## Why?

We want to just change scan pointer.

https://github.com/ruby/rexml/pull/114#discussion_r1501773803
> I want to just change scan pointer (`StringScanner#pos=`) instead of changing `@scanner.string`.

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -211,8 +211,9 @@ module REXML
         #STDERR.puts @source.encoding
         #STDERR.puts "BUFFER = #{@source.buffer.inspect}"
         if @document_status == nil
+          start_position = @source.position
           if @source.match("<?", true)
-            return process_instruction
+            return process_instruction(start_position)
           elsif @source.match("<!", true)
             if @source.match("--", true)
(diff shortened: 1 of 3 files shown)


lib/rexml/parsers/baseparser.rb
```diff
@@ -211,8 +211,9 @@ def pull_event
         #STDERR.puts @source.encoding
         #STDERR.puts "BUFFER = #{@source.buffer.inspect}"
         if @document_status == nil
+          pos = @source.pos
```
kou, line 214: How about renaming `pos` to `start_position` or something to indicate this is the position before we parsing the next event?

lib/rexml/source.rb
```diff
@@ -76,8 +76,12 @@ def match(pattern, cons=false)
       end
     end
 
-    def string=(string)
-      @scanner.string = string
+    def pos
+      @scanner.pos
+    end
+
+    def pos=(pos)
```
kou, line 83: How about avoid abbreviation here? The original word isn't so long.

```suggestion
    def position
      @scanner.pos
    end

    def position=(position)
```

kou:
Thanks!

PR #119, opened 2024-03-12
 Optimize the parse_attributes method to use `Source#match` to parse XML. 

naitoh:
## Why?

Improve maintainability by consolidating processing into `Source#match`.

## [Benchmark]
```
                 dom     10.891      10.622        16.356       17.403 i/s -     100.000 times in 9.182130s 9.414177s 6.113806s 5.746133s
                 sax     30.335      29.845        49.749       54.877 i/s -     100.000 times in 3.296483s 3.350595s 2.010071s 1.822259s
                pull     35.514      34.801        61.123       66.908 i/s -     100.000 times in 2.815793s 2.873484s 1.636041s 1.494591s
              stream     35.141      34.475        52.110       56.836 i/s -     100.000 times in 2.845646s 2.900638s 1.919017s 1.759456s

Comparison:
                              dom
         after(YJIT):        17.4 i/s
        before(YJIT):        16.4 i/s - 1.06x  slower
              before:        10.9 i/s - 1.60x  slower
               after:        10.6 i/s - 1.64x  slower

                              sax
         after(YJIT):        54.9 i/s
        before(YJIT):        49.7 i/s - 1.10x  slower
              before:        30.3 i/s - 1.81x  slower
               after:        29.8 i/s - 1.84x  slower

                             pull
         after(YJIT):        66.9 i/s
        before(YJIT):        61.1 i/s - 1.09x  slower
              before:        35.5 i/s - 1.88x  slower
               after:        34.8 i/s - 1.92x  slower

                           stream
         after(YJIT):        56.8 i/s
        before(YJIT):        52.1 i/s - 1.09x  slower
              before:        35.1 i/s - 1.62x  slower
               after:        34.5 i/s - 1.65x  slower

```

- YJIT=ON : 1.06x - 1.10x faster
- YJIT=OFF : 0.97x - 0.98x faster

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -114,7 +114,7 @@ module REXML
 
       module Private
         INSTRUCTION_END = /#{NAME}(\s+.*?)?\?>/um
-        TAG_PATTERN = /((?>#{QNAME_STR}))/um
+        TAG_PATTERN = /((?>#{QNAME_STR}))\s*/um
         CLOSE_PATTERN = /(#{QNAME_STR})\s*>/um
         ATTLISTDECL_END = /\s+#{NAME}(?:#{ATTDEF})*\s*>/um
         NAME_PATTERN = /\s*#{NAME}/um
@@ -128,7 +128,6 @@ module REXML
       def initialize( source )
         self.stream = source
         @listeners = []
-        @attributes_scanner = StringScanner.new('')
       end
 
       def add_listener( listener )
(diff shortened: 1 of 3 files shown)


lib/rexml/parsers/baseparser.rb
```diff
@@ -614,87 +613,60 @@ def process_instruction(start_position)
...
-            curr_ns << local_part
-          elsif prefix
-            prefixes << prefix unless prefix == "xml"
-          end
 
-          if attributes.has_key?(name)
-            msg = "Duplicate attribute #{name.inspect}"
-            raise REXML::ParseException.new(msg, @source, self)
-          end
+            if attributes[name]
```
kou, line 659: Why did you change `.has_key?` to `[]`?

test/parse/test_element.rb
```diff
@@ -39,11 +39,11 @@ def test_empty_namespace_attribute_name
           parse("<x :a=\"\"></x>")
         end
         assert_equal(<<-DETAIL.chomp, exception.to_s)
-Invalid attribute name: <:a="">
+Invalid attribute name: <:a=""></x>>
```
kou, line 42: Can we use `:a` as the invalid attribute name?

test/test_core.rb
```diff
@@ -1323,11 +1324,26 @@ def test_ticket_21
...
+      assert_equal(<<-DETAIL.chomp, exception.to_s)
 Missing attribute value start quote: <bar>
 Line: 1
 Position: 16
 Last 80 unconsumed characters:
+value/>
+      DETAIL
+    end
+
+    def test_ticket_21_part_2
```
kou, line 1336: Could you use meaningful test name?

The `ticket_21` refers the ticket ID of REXML's original bug trucker, which can't be referred now. :<

naitoh: OK.
I see.

I would change it to `test_parse_exception_on_missing_attribute_end_quote`.

lib/rexml/parsers/baseparser.rb
naitoh: I changed it to `attributes[name]` rather than `attributes.has_key?(name)` because it is faster to process and shorter to write.
Should I not have mixed it with this PR?

```ruby
$ cat benchmark/attributes.yaml 
loop_count: 100000
contexts:
  - name: No YJIT
    prelude: |
      require 'rexml'
  - name: YJIT
    prelude: |
      require 'rexml'
      RubyVM::YJIT.enable

prelude: |
  attributes = {}
  name = :a

benchmark:
  'attributes[name]' : attributes[name]
  'attributes.has_key?(name)' : attributes.has_key?(name)
```

```bash
$ benchmark-driver benchmark/attributes.yaml 
                             No YJIT        YJIT 
         attributes[name]    51.680M     54.437M i/s -    100.000k times in 0.001935s 0.001837s
attributes.has_key?(name)    46.019M     45.683M i/s -    100.000k times in 0.002173s 0.002189s

Comparison:
                      attributes[name]
                     YJIT:  54436580.8 i/s 
                  No YJIT:  51679591.0 i/s - 1.05x  slower

             attributes.has_key?(name)
                  No YJIT:  46019327.8 i/s 
                     YJIT:  45682957.4 i/s - 1.01x  slower
```

test/parse/test_element.rb
naitoh: It cannot be used as an attribute name, which raises an `Invalid attribute name` exception.
As a test case, what is the problem?

kou: I wanted to say that can we use the following as the expected value?

```suggestion
Invalid attribute name: <:a>
```

lib/rexml/parsers/baseparser.rb
kou: Oh... It's surprising...

Both of them just calls `hash_stlike_lookup()`:

`Hash#[]`: https://github.com/ruby/ruby/blob/b4f3f3c1031cc9ef5c6741042236db497be6602b/hash.c#L2135-L2146
`Hash#has_key?`: https://github.com/ruby/ruby/blob/b4f3f3c1031cc9ef5c6741042236db497be6602b/hash.c#L3671-L3675

Could you split this to another PR for now? I want to take a look at it later.

naitoh: OK. I See.

test/parse/test_element.rb
naitoh: It is difficult to determine the appropriate character for an attribute given invalid attributes such as the following.

- `'<x :a=""></x>'`
- `'<x =""></x>'`
- `'<x ""></x>'`

Since it is possible to determine the tag end, we will continue to return `:a=""` as an invalid tag as before.

Issue #25, opened 2020-02-21
Node set predicate not working


jcavalieri:
Hi,

I'm upgrading Ruby from 2.4.5 to 2.6.5 and the rexml behavior seems to have changed. We actually had some xpaths that were wrong that were working with the previous version. However, in correcting them I'm seeing this issue:

I expect the following example to return just one node.

```
xml_string = <<EOF
<div>
  <div>
    <test>1</test>
    <test>2</test>
  </div>
  <div>
    <test>1</test>
    <test>2</test>
  </div>
</div>
EOF

xml_doc = REXML::Document.new(xml_string)

nodes = REXML::XPath.match(xml_doc, '(/div/div/test[1])[1]')
```

It returns two nodes though.

```
>nodes.count
=> 2
```

```
> nodes.map(&:text)
=> ["1", "1"]
```

From my understanding placing an xpath subexpression in this `(...)[1]` should only return the first from the node set.

-John

jcavalieri:
I just test with 2.4.5 and I get the result that I expect.

```
> nodes.count
=> 1
```

jcavalieri:
It also appears to work with 2.5.

jcavalieri:
appears to be broken in 2.7 as well

kou:
Thanks for your report.
This is a bug. This should be fixed.

tompng:
This should be reopened. This is not fixed.

The root problem is in xpath parser. The two xpath `(/div/div/test[1])[1]` and `/div/div/test[1][1]` must give a different result, but the parsed xpath is the same. Perhaps the syntax tree lacks the ability to distinguish this two.
```ruby
with_paren = REXML::Parsers::XPathParser.new.parse('(/div/div/test[1])[1]')
without_paren = REXML::Parsers::XPathParser.new.parse('/div/div/test[1][1]')
with_paren == without_paren #=> true, should be false
```

```ruby
xml_string = <<EOF
<body>
<div>
  <div>
    <test>1</test>
    <test>2</test>
  </div>
  <div>
    <test>3</test>
    <test>4</test>
  </div>
</div>
<script>
// Reference: JavaScript's xpath
for (const xpath of ["//test[1][1]", "(//test[1])[1]"]) {
  snapshot = document.evaluate(xpath, document, null, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE, null);
[...]

PR #122, opened 2024-05-02
xpath: Fix wrong position with nested path

naitoh:
## Why?
Fixed incorrect calculation of position in node set.

See: https://github.com/ruby/rexml/issues/25


kou:
Thanks!

PR #139, opened 2024-06-05
Improve `Node#each_recursive` performance

### LINKED ISSUE #134: `each_recursive` is extremely slow
`REXML::Node#each_recursive` is a fundamental operation to traverse XML nodes. In particular, [CSS selector](https://github.com/makenowjust/rexml-css_selector) uses this method heavily.

Unfortunately, this method is extremely slow. The following Gist contains a tiny (but mostly equivalent) `each_recursive` implementation and their benchmarks.

<https://gist.github.com/makenowjust/b4852a59e53f0c85c740818c75303d5e>

And, the below is a result of this script on my laptop (Apple M1 Pro, 14 inch, 32 GB) and Ruby 3.3.2.

```console
$ ruby bench.rb
       user     system      total        real
   each_recursive  2.421999   0.008742   2.430741 (  2.453438)
my_each_recursive  0.057647   0.000049   0.057696 (  0.058718)
$ ruby --yjit bench.rb
       user     system      total        real
   each_recursive  1.613353   0.008227   1.621580 (  1.625374)
my_each_recursive  0.025493   0.000124   0.025617 (  0.025748)
```

Yes, REXML's `each_recursive` is ~50x (or ~80x with YJIT) slower than my tiny implementation.

I believe REXML does a lot of extra work, and we can make it faster.

makenowjust:
Fix #134

## Summary

This PR does:

- Add `benchmark/each_recursive.yaml`
- Rewrite `Node#each_recursive` implementation for performance
- Add a test for `Node#each_recursive`

The performance of `Node#each_recursive` is improved 60~80x faster.

## Details

`each_recursive` is too much slow as I described in #134.
I improved this performance by rewriting its implementation in this PR.

Also, I added a benchmark in `benchmark/each_recursive.yaml` and the following is a result on my laptop:

```
RUBYLIB= BUNDLER_ORIG_RUBYLIB= /Users/makenowjust/Projects/github.com/makenowjust/simple-dotfiles/.asdf/installs/ruby/3.3.2/bin/ruby -v -S benchmark-driver /Users/makenowjust/Projects/github.com/ruby/rexml/benchmark/each_recursive.yaml
ruby 3.3.2 (2024-05-30 revision e5a195edf6) [arm64-darwin23]
                     rexml 3.2.6      master  3.2.6(YJIT)  master(YJIT)
      each_recursive      11.279     686.502       17.926        1.470k i/s -     100.000 times in 8.866303s 0.145666s 5.578360s 0.068018s

Comparison:
                   each_recursive
        master(YJIT):      1470.2 i/s
              master:       686.5 i/s - 2.14x  slower
         3.2.6(YJIT):        17.9 i/s - 82.01x  slower
         rexml 3.2.6:        11.3 i/s - 130.35x  slower

```

We can see that the performance is improved 60~80x faster.

Additionally, I added a new test for `Node#each_recursive`.
It was missing, but we need it to confirm not to break the previous behavior.

Thank you.

### DIFF
--- a/lib/rexml/node.rb
+++ b/lib/rexml/node.rb
@@ -52,10 +52,14 @@ module REXML
 
     # Visit all subnodes of +self+ recursively
     def each_recursive(&block) # :yields: node
-      self.elements.each {|node|
-        block.call(node)
-        node.each_recursive(&block)
-      }
+      stack = []
+      each { |child| stack.unshift child if child.node_type == :element }
+      until stack.empty?
+        child = stack.pop
+        yield child
+        n = stack.size
+        child.each { |grandchild| stack.insert n, grandchild if grandchild.node_type == :element }
+      end
     end
 
     # Find (and return) first subnode (recursively) for which the block
--- a/test/test_document.rb
+++ b/test/test_document.rb
@@ -420,29 +420,29 @@ EOX
 
       class EachRecursiveTest < Test::Unit::TestCase
         def test_each_recursive
-          xml_source = <<-XML
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<root name="root">
-  <x name="0_2">
-    <x name="0_1">
-      <x name="0_0"/>
-    </x>
-  </x>
-  <x name="1_2">
-    <x name="1_1">
-      <x name="1_0"/>
-    </x>
-  </x>
-  <!-- comment -->
-  <![CDATA[ cdata ]]>
-</root>
-          XML
+          xml_source = <<~XML
+            <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+            <root name="root">
+              <x name="1_1">
+                <x name="1_2">
+                  <x name="1_3" />
+                </x>
+              </x>
+              <x name="2_1">
+                <x name="2_2">
+                  <x name="2_3" />
+                </x>
+              </x>
+              <!-- comment -->
+              <![CDATA[ cdata ]]>
+            </root>
+            XML
 
-          expected_names = [
-            "root",
-            "0_2", "0_1", "0_0",
-            "1_2", "1_1", "1_0",
-          ]         
+          expected_names = %w[
+            root
+            1_1 1_2 1_3
+            2_1 2_2 2_3
+          ]
 
           document = REXML::Document.new(xml_source)        
 
--- a/test/test_document.rb
+++ b/test/test_document.rb
@@ -421,13 +421,19 @@ EOX
       class EachRecursiveTest < Test::Unit::TestCase
         def test_each_recursive
           xml_source = +'<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
-          xml_source << '<root foo="bar">'
+          xml_source << '<root name="root">'
 
-          100.times do
+          expected_names = ["root"]
+
+          2.times do |x|
             x_node_source = ''
-            100.times do
-              x_node_source = "<x y='z'>#{x_node_source}</x>"
+            inner_names = []
+            3.times do |y|
+              name = "#{x}_#{y}"
+              x_node_source = "<x name='#{name}'>#{x_node_source}</x>"
+              inner_names << name
             end
+            expected_names.concat(inner_names.reverse)
             xml_source << x_node_source
           end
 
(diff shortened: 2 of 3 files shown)


lib/rexml/node.rb
```diff
@@ -52,10 +52,14 @@ def parent?
...
-        node.each_recursive(&block)
-      }
+      stack = []
+      each { |child| stack.unshift child if child.node_type == :element }
+      until stack.empty?
+        child = stack.pop
+        yield child
+        n = stack.size
+        child.each { |grandchild| stack.insert n, grandchild if grandchild.node_type == :element }
+      end
```
kou, line 62: Can we improve `REXML::Elements#each(nil)` case instead?

test/test_document.rb
```diff
@@ -417,6 +417,34 @@ def test_utf_16
...
       end
+
+      class EachRecursiveTest < Test::Unit::TestCase
+        def test_each_recursive
+          xml_source = +'<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
+          xml_source << '<root foo="bar">'
+
+          100.times do
+            x_node_source = ''
+            100.times do
```
kou, line 428: Do we need `100` for this test?
Can we reduce it?

Can we use different number for outer `times` and inner `times` for easy to distinct a child problem and a descendant problem.

test/test_document.rb
```diff
@@ -417,6 +417,34 @@ def test_utf_16
...
+
+          xml_source << '</root>'
+
+          document = REXML::Document.new(xml_source)        
+
+          count = 0
+          document.each_recursive { count += 1 }
+          # Node#each_recursive iterates elements only.
+          # This does not iterate XML declerations, comments, attributes, CDATA sections, etc.
+          assert_equal(100 * 100 + 1, count)
```
kou, line 445: Can we use all collected element name instead of just the number of loop count?
Count only check will be difficult to debug when this is failed.

lib/rexml/node.rb
makenowjust: This seems a possible improvement. But, IMHO, it can be split into another pull request.

kou: Do we still need to change `REXML::Node#each_recursive` when we improve `REXML::Elements#each(nil)`?

test/test_document.rb
```diff
@@ -417,6 +417,42 @@ def test_utf_16
...
+              inner_names << name
+            end
+            expected_names.concat(inner_names.reverse)
+            xml_source << x_node_source
+          end
+
+          xml_source << '<!-- comment -->'
+          xml_source << '<![CDATA[ cdata ]]>'
+
+          xml_source << '</root>'
```
kou, line 443: Can we use literal for this test data for readability something like the following?

```suggestion
          xml_source = <<-XML
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<root name="root">
  <x name="0_2">
    <x name="0_1">
      <x name="0_0"/>
    </x>
  </x>
  <x name="1_2">
    <x name="1_1">
      <x name="1_0"/>
    </x>
  </x>
  <!-- comment -->
  <![CDATA[ cdata ]]>
</root>
          XML

          expected_names = [
            "root",
            "0_2", "0_1", "0_0",
            "1_2", "1_1", "1_0",
          ]         
```

makenowjust: I have no opinion.

lib/rexml/node.rb
makenowjust: This implementation is best so like that we don't need to change it when we improve something.

I don't really like implementations that are optimized when using specific arguments. From that point of view, I'm not very keen on optimizing `each(nil)`.

makenowjust: If `elements.each(nil)` is improved, we can use `elements.each { |x| yield x }` instead of `each { |x| yield x if x.node_type == :element }`.

However, I don't think it is a problem that there are some duplications of `each { |x| yield x if x.node_type == :element }` sometimes. Because `elements.each(&)` has some overheads and the performance is more important here.

Repeatedly, I talked about the **performance**. The maintainability and other matters are out of topic here because the speed of `each_recursive` is one of the crucial parts of the XML library.

test/test_document.rb
```diff
@@ -417,6 +417,44 @@ def test_utf_16
           assert_equal(expected_xml, actual_xml)
         end
       end
+
+      class EachRecursiveTest < Test::Unit::TestCase
+        def test_each_recursive
```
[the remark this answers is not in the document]
makenowjust: Okay. Thank you.

test/test_document.rb
```diff
@@ -417,6 +417,44 @@ def test_utf_16
...
+              </x>
+              <x name="2_1">
+                <x name="2_2">
+                  <x name="2_3" />
+                </x>
+              </x>
+              <!-- comment -->
+              <![CDATA[ cdata ]]>
+            </root>
+            XML
```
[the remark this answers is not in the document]
makenowjust: Personally, I'd like to keep the same indentations of heredoc end marker and contents. But, I have no strong opinion.

lib/rexml/node.rb
makenowjust: Thank you for your understanding!

kou:
I've merged.
Thanks.

PR #144, opened 2024-06-11
Remove an unused class var `@@namespaces`

makenowjust:
`@@namespaces` is defined under `REXML`, but it is never used.
At least, `rake test` passes when it is removed.

I guess the comment above `@@namespaces` is also false.


kou:
Good catch.

PR #145, opened 2024-06-12
Add a "malformed comment" check for top-level comments

makenowjust:
This check was missing. Therefore, `REXML::Document.new("<!--")` raised the ``undefined method `[]' for nil`` error, for example.

This PR also adds tests for "malformed comment" checks.

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -228,7 +228,14 @@ module REXML
             return process_instruction(start_position)
           elsif @source.match("<!", true)
             if @source.match("--", true)
-              return [ :comment, @source.match(/(.*?)-->/um, true)[1] ]
+              md = @source.match(/(.*?)-->/um, true)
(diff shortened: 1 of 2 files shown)


lib/rexml/parsers/baseparser.rb
```diff
@@ -228,7 +228,14 @@ def pull_event
             return process_instruction(start_position)
           elsif @source.match("<!", true)
             if @source.match("--", true)
-              return [ :comment, @source.match(/(.*?)-->/um, true)[1] ]
+              md = @source.match(/(.*?)-->/um, true)
+              if md.nil?
+                raise REXML::ParseException.new("Unclosed comment", @source)
+              end
+              if md[1] =~ /--|-\z/
```
kou, line 235: `Regexp#match?` will be faster:

```suggestion
              if /--|-\z/.match?(md[1])
```

kou:
Good catch.

PR #147, opened 2024-06-13
Add a "Malformed comment" check for invalid comments such as `<!-->`

makenowjust:
`Document.new("<a><!-->")` raises `undefined method '[]' for nil`. This commit fixes it and adds a test for it.


olleolleolle:
See #145 for the same sort of check.

I like it!

kou:
Good catch.

PR #148, opened 2024-06-13
Fix small typos

makenowjust:
I found these typos with using [`typos-cli`](https://github.com/crate-ci/typos).

Now, we can obtain no typo reports from the `typos` command with this configuration (`.typos.toml`):

```toml
[files]
extend-exclude = [
  "*.svg",
  "*.xml",
]

[default.extend-words]
# Common variable names in this project.
arry = "arry"
blok = "blok"
eles = "eles"
# Incomplete words in test data.
caf = "caf"
# German words in test data.
abl = "abl" # NOTE: It is a part of "Ablüfe".
alle = "alle"
ist = "ist"
technik = "technik"
```

Thank you.

PR #149, opened 2024-06-13
Reinstate "SER" in NOTE comment [ci skip]

olleolleolle:
This is a follow-up to a previous change today that I had misunderstood, this puts the author SER back into the note.

See #148


makenowjust:
Thank you @olleolleolle.

kou:
Ah, we don't need this because we're maintaining REXML by community not only the original author.
We should share NOTEs with maintainers and contributors.

PR #151, opened 2024-06-18
Handle unprefixed attributes in `Element#attribute` correctly

makenowjust:
Currently, `Element#attrubute` handles unprefixed attributes as attributes with the default namespace, and handles the default namespace as `nil` value. However, it is a mis-interpretation of the specification of XML 1.0.

XML 1.0 specification says:
(https://www.w3.org/TR/xml-names/#defaulting)

> The namespace name for an unprefixed attribute name always has no value.

Therefore, we need to distinguish `nil` and the default namespace. Also, unprefixed attributes should not be handled as attributes with the default namespace.

Note that XML 1.0 specification also says:

> Default namespace declarations do not apply directly to attribute names; the interpretation of unprefixed attributes is determined by the element on which they appear.

This means that we can know (programmatically) how to handle unprefixed attributes from the element's namespace. It does not mean that the namespace of an unprefixed attribute becomes the default namespace.

This commit changes `Element#attribute` behavior correctly and updates tests for it.

- - -

@kou It breaks backward compatibility, so I'm concerned. What do you think?

### DIFF
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -1270,27 +1270,14 @@ module REXML
     # With arguments +name+ and +namespace+ given,
     # returns the value of the named attribute if it exists, otherwise +nil+:
     #
-    #   xml_string = "<root xmlns:a='a' a:x='a:x' x='x'/>"
+    #   xml_string = "<root xmlns:a='http://example.com/a' a:x='a:x' x='x'/>"
     #   document = REXML::Document.new(xml_string)
     #   document.root.attribute("x")      # => x='x'
-    #   document.root.attribute("x", "a") # => a:x='a:x'
+    #   document.root.attribute("x", "http://example.com/a") # => a:x='a:x'
     #
     def attribute( name, namespace=nil )
       prefix = namespaces.key(namespace) if namespace
-      prefix = nil if prefix == 'xmlns'
-
-      ret_val =
-        attributes.get_attribute( prefix ? "#{prefix}:#{name}" : name )
-
-      return ret_val unless ret_val.nil?
-      return nil if prefix.nil?
-
-      # now check that prefix'es namespace is not the same as the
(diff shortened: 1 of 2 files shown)


lib/rexml/element.rb
```diff
@@ -1270,27 +1270,14 @@ def [](name_or_index)
     # With arguments +name+ and +namespace+ given,
     # returns the value of the named attribute if it exists, otherwise +nil+:
     #
-    #   xml_string = "<root xmlns:a='a' a:x='a:x' x='x'/>"
+    #   xml_string = "<root xmlns:a='http://example.com/a' a:x='a:x' x='x'/>"
     #   document = REXML::Document.new(xml_string)
     #   document.root.attribute("x")      # => x='x'
-    #   document.root.attribute("x", "a") # => a:x='a:x'
+    #   document.root.attribute("x", "http://example.com/a") # => a:x='a:x'
```
makenowjust, line 1276: I also updated here to make clear that the `namespace` argument takes a namespace URI.

makenowjust:
@kou I agree with deprecating `Element#attribute`. Also, I'm not sure a new strict version is necessary because we have `Element#attributes` and `Attributes#get_attribute`.

kou:
Then how about documenting this instead of changing the current behavior? 
`Element#attributes` and `Attributes#get_attribute` will be too long as an alternative way. We can revisit an API for this when someone claims this again.

makenowjust:
In many cases, people want to obtain an attribute value, and don't want an `Attribute` object. Thus, people should use `element.attributes[name]`, and long `elemene.attributes.get_attributes(name)` is not a problem. (`get_attribute` seems redundant to me, but it is another problem.)

kou:
> `element.attributes[name]`

FYI: We can use `element[name]`. :-)

PR #153, opened 2024-06-18
Reject unclosed DOCTYPE on parsing

### LINKED ISSUE #152: BUG: REXML accepts unclosed DOCTYPE
Currently, REXML accepts unclosed DOCTYPE declarations.

```ruby
require "rexml/document"

REXML::Document.new("<!DOCTYPE foo [")
# => <UNDEFINED> ... </>
# No exceptions are raised!
```

It seems strange and it should be fixed.

makenowjust:
Fix #152

### DIFF
--- a/lib/rexml/parsers/treeparser.rb
+++ b/lib/rexml/parsers/treeparser.rb
@@ -16,7 +16,6 @@ module REXML
 
       def parse
         tag_stack = []
-        in_doctype = false
         entities = nil
         begin
           while true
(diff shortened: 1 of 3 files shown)


test/parse/test_document_type_declaration.rb
```diff
@@ -53,6 +53,53 @@ def test_no_name
       end
     end
 
+    class TestUnclosed < self
+      def test_unclosed_doctype_only
```
kou, line 57: `unclosed_doctype_` is redundant because file name is `test_document_type_declaration.rb` and class name is `TestUnclosed`.

```suggestion
      def test_no_extra_node
```

lib/rexml/parsers/treeparser.rb
```diff
@@ -24,12 +24,20 @@ def parse
             #STDERR.puts "TREEPARSER GOT #{event.inspect}"
             case event[0]
             when :end_document
+              if in_doctype
+                raise ParseException.new("Malformed DOCTYPE: unclosed",
+                                         @parser.source, @parser)
+              end
```
kou, line 30: Can we do this in `baseparser.rb` instead of here?
If we can do it, other parsers such as `StreamParser` can detect this case.

makenowjust: I will try it.

makenowjust: Updated!

Issue #150, opened 2024-06-17
Can't parse large XML


kou:
For example, we can't parse an XML of Wikipedia: https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2

```text
REXML::ParseException: #<RangeError: integer 2147501889 too big to convert to 'int'>
/tmp/local/lib/ruby/gems/3.4.0+0/gems/rexml-3.3.0/lib/rexml/source.rb:127:in 'StringScanner#pos='
/tmp/local/lib/ruby/gems/3.4.0+0/gems/rexml-3.3.0/lib/rexml/source.rb:127:in 'REXML::Source#position='
/tmp/local/lib/ruby/gems/3.4.0+0/gems/rexml-3.3.0/lib/rexml/parsers/baseparser.rb:447:in 'REXML::Parsers::BaseParser#pull_event'
/tmp/local/lib/ruby/gems/3.4.0+0/gems/rexml-3.3.0/lib/rexml/parsers/baseparser.rb:207:in 'REXML::Parsers::BaseParser#pull'
/tmp/local/lib/ruby/gems/3.4.0+0/gems/rexml-3.3.0/lib/rexml/parsers/streamparser.rb:20:in 'REXML::Parsers::StreamParser#parse'
/tmp/local/lib/ruby/gems/3.4.0+0/gems/red-datasets-0.1.7/lib/datasets/wikipedia.rb:51:in 'block in Datasets::Wikipedia#each'
/tmp/local/lib/ruby/gems/3.4.0+0/gems/red-datasets-0.1.7/lib/datasets/dataset.rb:78:in 'block (2 levels) in Datasets::Dataset#extract_bz2'
/tmp/local/lib/ruby/gems/3.4.0+0/gems/red-datasets-0.1.7/lib/datasets/dataset.rb:56:in 'IO.pipe'
/tmp/local/lib/ruby/gems/3.4.0+0/gems/red-datasets-0.1.7/lib/datasets/dataset.rb:56:in 'block in Datasets::Dataset#extract_bz2'
/tmp/local/lib/ruby/gems/3.4.0+0/gems/red-datasets-0.1.7/lib/datasets/dataset.rb:55:in 'IO.pipe'
/tmp/local/lib/ruby/gems/3.4.0+0/gems/red-datasets-0.1.7/lib/datasets/dataset.rb:55:in 'Datasets::Dataset#extract_bz2'
[...]

PR #154, opened 2024-06-19
Fix a bug that a large XML can't be parsed

naitoh:
GitHub: fix GH-150

If a parsed XML is later than `2 ** 31 - 1`, we can't parse it. Because `StringScanner`s position is stored as `int`. We can avoid the restriction by dropping large parsed content.

### DIFF
--- a/lib/rexml/source.rb
+++ b/lib/rexml/source.rb
@@ -55,6 +55,7 @@ module REXML
     attr_reader :encoding
 
     module Private
+      SCANNER_RESET_SIZE = 100000
       PRE_DEFINED_TERM_PATTERNS = {}
       pre_defined_terms = ["'", '"', "<"]
       pre_defined_terms.each do |term|
@@ -82,6 +83,12 @@ module REXML
     # The current buffer (what we're going to read next)
     def buffer
       @scanner.rest
+    end
+
+    def drop_parsed_content
+      if @scanner.pos > Private::SCANNER_RESET_SIZE
+        @scanner.string = @scanner.rest
+      end
     end
 
     def buffer_encoding=(encoding)
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -204,6 +204,8 @@ module REXML
 
       # Returns the next event.  This is a +PullEvent+ object.
       def pull
+        @source.drop_parsed_content
+
         pull_event.tap do |event|
           @listeners.each do |listener|
             listener.receive event
--- a/test/parser/test_base_parser.rb
+++ b/test/parser/test_base_parser.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: false
+
+require 'rexml/parsers/baseparser'
+
+module REXMLTests
+  class BaseParserTester < Test::Unit::TestCase
+    def test_large_xml
+      large_text = "a" * 100_000
+      xml = <<-XML
+        <?xml version="1.0"?>
+        <root>
+          <child>#{large_text}</child>
+          <child>#{large_text}</child>
+        </root>
+      XML
+
+      parser = REXML::Parsers::BaseParser.new(xml)
+      while parser.has_next?
+        parser.pull
+      end
+
+      assert do
+        parser.position < xml.bytesize
+      end
+    end
+  end
+end


lib/rexml/source.rb
```diff
@@ -55,6 +55,7 @@ class Source
     attr_reader :encoding
 
     module Private
+      SCANNER_RESET_SIZE = 100000
```
kou, line 58: How about making this customizable only in test and using small XML in test for faster test?

test/test_pullparser.rb
```diff
@@ -98,5 +98,28 @@ def test_peek
...
+        xml <<  '</child>'
+      end
+      xml << '</root>'
+    end
+
+    # NOTE: this test is too slow.
+    def test_parse_large_xml
+      xml = build_xml(N_ELEMENTS)
+
+      parser = REXML::Parsers::PullParser.new(xml)
```
kou, line 119: How about using `BaseParser` and `Source` instead of `PullParser` and `String`?
Then we can use `Source#position` for assertion.

naitoh: OK, I see.
Thanks!

lib/rexml/source.rb
naitoh: This one was improved even with fixed values.
Thanks!

test/test_baseparser.rb
kou, line 1: Could you move this to `test/parser/test_base_parser.rb`?

test/test_baseparser.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: false
+
+require 'rexml/parsers/baseparser'
+
+module REXMLTests
+  class BaseParserTester < Test::Unit::TestCase
+    include REXML
```
kou, line 7: This is needless.

```suggestion
```

test/test_baseparser.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    def test_parse_large_xml
+      xml = build_xml(N_ELEMENTS)
+      parser = REXML::Parsers::BaseParser.new(xml)
+      while parser.has_next?
+        parser.pull
+        assert_compare REXML::Source::SCANNER_RESET_SIZE + N_STRING.size, ">", parser.position
+      end
+    end
```
kou, line 29: How about doing them all in the test because `build_xml` will not be reused? Something like:

```ruby
def test_large_xml
  large_text = "a" * 100_000
  xml = <<-XML
    <?xml version="1.0"?>
    <root>
      <child>#{large_text}</child>
      <child>#{large_text}</child>
    </root>
  XML

  parser = REXML::Parsers::BaseParser.new(xml)
  while parser.has_next?
    parser.pull
  end
  assert do
    source.position < xml.bytesize
  end
end
```

test/test_baseparser.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+      xml << '</root>'
+    end
+
+    def test_parse_large_xml
+      xml = build_xml(N_ELEMENTS)
+      parser = REXML::Parsers::BaseParser.new(xml)
+      while parser.has_next?
+        parser.pull
+        assert_compare REXML::Source::SCANNER_RESET_SIZE + N_STRING.size, ">", parser.position
```
kou, line 27: !!! Can we access `SCANNER_REST_SIZE` via ` REXML::Source` even when we use `private_constant`!?

test/test_baseparser.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: false
+
+require 'rexml/parsers/baseparser'
+
+module REXMLTests
+  class BaseParserTester < Test::Unit::TestCase
+    include REXML
+
+    N_ELEMENTS = 100
+    N_STRING = 'a' * 50000
```
kou, line 10: This name isn't good because `N_` means "the number of ".

test/test_baseparser.rb
naitoh: OK
I see.

test/test_baseparser.rb
naitoh: OK
I see.

test/test_baseparser.rb
naitoh: Yes.
```
    20| require 'debug'
    21|     def test_parse_large_xml
    22|       xml = build_xml(N_ELEMENTS)
    23|       parser = REXML::Parsers::BaseParser.new(xml)
=>  24|       debugger
    25|       while parser.has_next?
    26|         parser.pull
    27|         assert_compare REXML::Source::SCANNER_RESET_SIZE + N_STRING.size, ">", parser.position
    28|       end
=>#0	REXMLTests::BaseParserTester#test_parse_large_xml at ~/ghq/github.com/naitoh/rexml/test/parser/test_base.rb:24
  #1	Test::Unit::TestCase#run_test at ~/.rbenv/versions/3.3.2/lib/ruby/gems/3.3.0/gems/test-unit-3.6.1/lib/test/unit/testcase.rb:871
  # and 28 frames (use `bt' command for all frames)
(ruby) puts REXML::Source::SCANNER_RESET_SIZE
100000
nil

```

test/test_baseparser.rb
naitoh: Oh, yes, indeed...
I understand.

test/test_baseparser.rb
kou: Oh... It seems that we should remove `include Private` (in a separated PR)...

naitoh: Yes...

test/test_baseparser.rb
naitoh: OK.
I See.

kou:
Thanks.

PR #155, opened 2024-06-22
Don't include private_constant-ed module

naitoh:
Included constants are not private. So private constants in private module aren't private.

See also: https://github.com/ruby/rexml/pull/154#discussion_r1649469269


kou:
Thanks.

PR #161, opened 2024-07-04
fix: Extra content at the end of the document

naitoh:
## Why?

XML with additional content at the end of the document is invalid.

https://www.w3.org/TR/2006/REC-xml11-20060816/#document

```
[1]   document   ::=   ( prolog element Misc* ) - ( Char* RestrictedChar Char* )
```

https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-Misc

```
[27]    Misc       ::=          Comment | PI | S
```

https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-PI

```
[16]    PI         ::=          '<?' PITarget (S (Char* - (Char* '?>' Char*)))? '?>'
```

https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-PITarget

```
[17]    PITarget           ::=          Name - (('X' | 'x') ('M' | 'm') ('L' | 'l'))
```

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -460,14 +460,23 @@ module REXML
                 @closed = tag
                 @nsstack.shift
               else
+                if @tags.empty? and @have_root
+                  raise ParseException.new("Malformed XML: Extra tag at the end of the document (got '<#{tag}')", @source)
+                end
                 @tags.push( tag )
               end
+              @have_root = true
               return [ :start_element, tag, attributes ]
             end
           else
             text = @source.read_until("<")
             if text.chomp!("<")
               @source.position -= "<".bytesize
+            end
+            if @tags.empty? and @have_root
+              unless /\A\s*\z/.match?(text)
+                raise ParseException.new("Malformed XML: Extra content at the end of the document (got '#{text}')", @source)
+              end
             end
             return [ :text, text ]
           end
--- a/test/parse/test_comment.rb
+++ b/test/parse/test_comment.rb
@@ -105,5 +105,17 @@ module REXMLTests
         DETAIL
       end
     end
+
+    def test_after_root
+      parser = REXML::Parsers::BaseParser.new('<a></a><!-- ok comment -->')
+
+      events = {}
+      while parser.has_next?
+        event = parser.pull
+        events[event[0]] = event[1]
+      end
+
+      assert_equal(" ok comment ", events[:comment])
+    end
   end
 end
--- a/test/parse/test_element.rb
+++ b/test/parse/test_element.rb
@@ -85,6 +85,40 @@ Last 80 unconsumed characters:
 </ </x>
         DETAIL
       end
+
+      def test_after_root
+        exception = assert_raise(REXML::ParseException) do
+          parser = REXML::Parsers::BaseParser.new('<a></a><b>')
+          while parser.has_next?
+            parser.pull
+          end
+        end
+
+        assert_equal(<<~DETAIL.chomp, exception.to_s)
+          Malformed XML: Extra tag at the end of the document (got '<b')
+          Line: 1
+          Position: 10
+          Last 80 unconsumed characters:
+
+        DETAIL
+      end
+
+      def test_after_empty_element_tag_root
+        exception = assert_raise(REXML::ParseException) do
+          parser = REXML::Parsers::BaseParser.new('<a/><b>')
+          while parser.has_next?
+            parser.pull
+          end
+        end
+
+        assert_equal(<<~DETAIL.chomp, exception.to_s)
+          Malformed XML: Extra tag at the end of the document (got '<b')
+          Line: 1
+          Position: 7
+          Last 80 unconsumed characters:
+
+        DETAIL
+      end
     end
   end
 end
--- a/test/parse/test_text.rb
+++ b/test/parse/test_text.rb
@@ -1,0 +1,25 @@
+require "test/unit"
+require 'rexml/parsers/baseparser'
+
+module REXMLTests
+  class TestParseText < Test::Unit::TestCase
+    class TestInvalid < self
+      def test_after_root
+        exception = assert_raise(REXML::ParseException) do
+          parser = REXML::Parsers::BaseParser.new('<a></a>c')
+          while parser.has_next?
+            parser.pull
+          end
+        end
+
+        assert_equal(<<~DETAIL.chomp, exception.to_s)
+          Malformed XML: Extra content at the end of the document (got 'c')
+          Line: 1
+          Position: 8
+          Last 80 unconsumed characters:
+
+        DETAIL
+      end
+    end
+  end
+end
--- a/test/parse/test_processing_instruction.rb
+++ b/test/parse/test_processing_instruction.rb
@@ -40,5 +40,17 @@ x<?x y
                      ])
       end
     end
+
+    def test_after_root
+      parser = REXML::Parsers::BaseParser.new('<a></a><?abc version="1.0" ?>')
+
+      events = {}
+      while parser.has_next?
+        event = parser.pull
+        events[event[0]] = event[1]
+      end
+
+      assert_equal("abc", events[:processing_instruction])
+    end
   end
 end
--- a/test/test_pullparser.rb
+++ b/test/test_pullparser.rb
@@ -63,8 +63,10 @@ module REXMLTests
     end
 
     def test_character_references
-      source = '<a>&#65;</a><b>&#x42;</b>'
+      source = '<root><a>&#65;</a><b>&#x42;</b></root>'
       parser = REXML::Parsers::PullParser.new( source )
+
+      events = {}
       element_name = ''
       while parser.has_next?
         event = parser.pull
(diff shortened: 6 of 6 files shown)


lib/rexml/parsers/baseparser.rb
```diff
@@ -460,15 +461,24 @@ def pull_event
                 @closed = tag
                 @nsstack.shift
               else
+                if @tags.size.zero? && !@root_tag.nil?
+                  raise ParseException.new("Malformed XML: Extra tag at the end of the document (got '<#{tag}')", @source)
+                end
                 @tags.push( tag )
               end
+              @root_tag ||= tag
```
kou, line 469: How about using boolean for root element check because we don't need root element name?

```suggestion
                if @tags.empty?
                  if @have_root
                    raise ParseException.new("Malformed XML: Extra tag at the end of the document (got '<#{tag}')", @source)
                  end
                  @have_root = true
                end
                @tags.push( tag )
              end
```

lib/rexml/parsers/baseparser.rb
```diff
@@ -460,15 +461,24 @@ def pull_event
...
               return [ :start_element, tag, attributes ]
             end
           else
             text = @source.read_until("<")
             if text.chomp!("<")
               @source.position -= "<".bytesize
             end
+            if @tags.size.zero? && !@root_tag.nil?
```
kou, line 477: * Could you reduce method calls?
* Could you use `and` not `&&`?

```suggestion
            if @tags.empty? and !@root_tag.nil?
```

lib/rexml/parsers/baseparser.rb
```diff
@@ -635,6 +645,9 @@ def process_instruction(start_position)
           @source.position = start_position
           raise REXML::ParseException.new(message, @source)
         end
+        if @tags.size.zero? && !@root_tag.nil? && match_data[1] == "xml"
+          raise ParseException.new("Malformed XML: Extra XML declaration at the end of the document (got '<?#{match_data[1]}')", @source)
+        end
```
kou, line 650: XML declaration must be the first item. So we can use different condition for the XML declaration position check.
For example:

```ruby
if match_data[1] == "xml"
  raise ... if @document_status
  # ...
end
```

How about working on this in a separated PR?

test/parser/test_base_parser.rb
```diff
@@ -23,5 +23,80 @@ def test_large_xml
         parser.position < xml.bytesize
       end
     end
+
+    def test_parse_exception_for_multiple_root_elements
```
kou, line 27: Could you move this to https://github.com/ruby/rexml/blob/master/test/parse/test_element.rb ?

test/parser/test_base_parser.rb
```diff
@@ -23,5 +23,80 @@ def test_large_xml
...
+      assert_equal(<<-DETAIL.chomp, exception.to_s)
+Malformed XML: Extra tag at the end of the document (got '<b')
+Line: 1
+Position: 10
+Last 80 unconsumed characters:
+
+      DETAIL
+    end
+
+    def test_parse_exception_for_extra_content_at_the_end_of_the_document
```
kou, line 44: Could you move this to https://github.com/ruby/rexml/blob/master/test/parse/test_text.rb (the file doesn't exist yet...)?

test/parser/test_base_parser.rb
```diff
@@ -23,5 +23,80 @@ def test_large_xml
...
+      assert_equal(<<-DETAIL.chomp, exception.to_s)
+Malformed XML: Extra XML declaration at the end of the document (got '<?xml')
+Line: 1
+Position: 29
+Last 80 unconsumed characters:
+
+      DETAIL
+    end
+
+    def test_extra_processing_instruction_node_at_the_end_of_the_document
```
kou, line 78: Could you move this to https://github.com/ruby/rexml/blob/master/test/parse/test_processing_instruction.rb ?

test/parser/test_base_parser.rb
```diff
@@ -23,5 +23,80 @@ def test_large_xml
...
+      events = {}
+      while parser.has_next?
+        event = parser.pull
+        events[event[0]] = event[1]
+      end
+
+      assert_equal("abc", events[:processing_instruction])
+    end
+
+    def test_extra_comments_at_the_end_of_the_document
```
kou, line 90: Could you move this to https://github.com/ruby/rexml/blob/master/test/parse/test_comment.rb ?

test/parser/test_base_parser.rb
naitoh: I see.
OK.

test/parser/test_base_parser.rb
naitoh: OK.
I see.

test/parser/test_base_parser.rb
naitoh: OK.
I see.

test/parser/test_base_parser.rb
naitoh: OK.
I see.

lib/rexml/parsers/baseparser.rb
naitoh: OK.
I see.

lib/rexml/parsers/baseparser.rb
naitoh: OK.
I see.
Thanks!

lib/rexml/parsers/baseparser.rb
naitoh: OK.
I see.
I will make another PR.
Thanks!

naitoh: I created https://github.com/ruby/rexml/pull/162

test/parse/test_element.rb
```diff
@@ -85,6 +85,40 @@ def test_garbage_less_than_slash_before_end_tag_at_line_start
...
+        assert_equal(<<~DETAIL.chomp, exception.to_s)
+          Malformed XML: Extra tag at the end of the document (got '<b')
+          Line: 1
+          Position: 10
+          Last 80 unconsumed characters:
+
+        DETAIL
+      end
+
+      def test_multiple_root_elements_with_self_close_tag
```
kou, line 106:
```suggestion
      def test_after_empty_element_tag_root
```

test/parse/test_text.rb
```diff
@@ -0,0 +1,25 @@
+require "test/unit"
+require 'rexml/parsers/baseparser'
+
+module REXMLTests
+  class TestParseText < Test::Unit::TestCase
+    class TestInvalid < self
+      def test_extra_content_at_the_end_of_the_document
```
kou, line 7:
```suggestion
      def test_after_root
```

lib/rexml/parsers/baseparser.rb
```diff
@@ -460,15 +460,24 @@ def pull_event
...
+              @have_root = true
               return [ :start_element, tag, attributes ]
             end
           else
             text = @source.read_until("<")
             if text.chomp!("<")
               @source.position -= "<".bytesize
             end
+            if @tags.empty? and @have_root
+              if text.strip != ""
```
kou, line 477: `strip` allocates a new string. Can we avoid it?
For example: `/\A\s*\z/.match?(text)`

test/parse/test_comment.rb
```diff
@@ -105,5 +105,17 @@ def test_after_doctype_malformed_comment_end
         DETAIL
       end
     end
+
+    def test_extra_comments_at_the_end_of_the_document
```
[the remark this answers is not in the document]
naitoh: OK.
I see.

test/parse/test_element.rb
```diff
@@ -85,6 +85,40 @@ def test_garbage_less_than_slash_before_end_tag_at_line_start
 </ </x>
         DETAIL
       end
+
+      def test_multiple_root_elements
```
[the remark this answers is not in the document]
naitoh: OK.
I see.

test/parse/test_element.rb
naitoh: OK.
I see.

test/parse/test_text.rb
naitoh: OK.
I see.

kou:
Thanks.

DmitryPogrebnoy:
@naitoh @kou After this change, parsing of `'<threads><thread id="1"/></threads><threads><thread id="1"/></threads>'` using `REXML::Parsers::PullParser.new(socket)` throw an illegal seek exception. Should I create a separate issue for this problem?

The main cause is this if statement at lib/rexml/parsers/baseparser.rb:498
```
if @tags.empty? and @have_root
  raise ParseException.new("Malformed XML: Extra tag at the end of the document (got '<#{tag}')", @source)
end
```

kou:
`<threads><thread id="1"/></threads><threads><thread id="1"/></threads>` is an invalid XML.
Are you suggesting that we should support an invalid XML?

kou:
OK. Could you open a new issue for it?

DmitryPogrebnoy:
@kou Yep, here it is - https://github.com/ruby/rexml/issues/214

PR #162, opened 2024-07-06
Add position check for XML declaration

naitoh:
## Why?
XML declaration must be the first item.

https://www.w3.org/TR/2006/REC-xml11-20060816/#document

```
[1]   document   ::=   ( prolog element Misc* ) - ( Char* RestrictedChar Char* )
```

https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-prolog

```
[22]   prolog   ::=   	XMLDecl Misc* (doctypedecl Misc*)?
```

https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-XMLDecl

```
[23]   XMLDecl  ::=   '<?xml' VersionInfo EncodingDecl? SDDecl? S? '?>'
```

See: https://github.com/ruby/rexml/pull/161#discussion_r1666118193

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -635,7 +635,10 @@ module REXML
           @source.position = start_position
           raise REXML::ParseException.new(message, @source)
         end
-        if @document_status.nil? and match_data[1] == "xml"
+        if match_data[1] == "xml"
+          if @document_status
+            raise ParseException.new("Malformed XML: XML declaration is not at the start", @source)
+          end
           content = match_data[2]
           version = VERSION.match(content)
           version = version[1] unless version.nil?
--- a/test/parse/test_processing_instruction.rb
+++ b/test/parse/test_processing_instruction.rb
@@ -39,6 +39,23 @@ x<?x y
                        pi.content,
                      ])
       end
+
+      def test_xml_declaration_not_at_document_start
+        exception = assert_raise(REXML::ParseException) do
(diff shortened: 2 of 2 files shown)


lib/rexml/parsers/baseparser.rb
```diff
@@ -635,21 +635,25 @@ def process_instruction(start_position)
...
-          encoding = encoding[1] unless encoding.nil?
-          if need_source_encoding_update?(encoding)
-            @source.encoding = encoding
-          end
-          if encoding.nil? and /\AUTF-16(?:BE|LE)\z/i =~ @source.encoding
-            encoding = "UTF-16"
+        if match_data[1] == "xml"
+          if @document_status
+            raise ParseException.new("Malformed XML: XML declaration other than at the top of the document.", @source)
+          else
```
kou, line 641: * How about early return style to reduce indent?
* Could you remove `.` in the error message?
* Can we simplify the error message?

```suggestion
          unless @document_status
            raise ParseException.new("Malformed XML: XML declaration is not at the start", @source)
          end
```

naitoh: `unless @document_status` would result in the opposite branch.
The correct answer is `if @document_status`, right?

```
          if @document_status
            raise ParseException.new("Malformed XML: XML declaration is not at the start", @source)
          end
```

kou: Oh, sorry.

kou:
Thanks.

PR #164, opened 2024-07-09
Stop adding extra new line after XML declaration with pretty format

naitoh:
If the XML file does not end with a newline, a space is added to the end of the first line.

```ruby
Failure: test_indent(REXMLTests::TestDocument::WriteTest::ArgumentsTest)
/Users/naitoh/ghq/github.com/naitoh/rexml/test/test_document.rb:270:in `test_indent'
     267:           output = ""
     268:           indent = 2
     269:           @document.write(output, indent)
  => 270:           assert_equal(<<-EOX.chomp, output)
     271: <?xml version='1.0' encoding='UTF-8'?>
     272: <message>
     273:   Hello world!
<"<?xml version='1.0' encoding='UTF-8'?>\n" +
"<message>\n" +
"  Hello world!\n" +
"</message>"> expected but was
<"<?xml version='1.0' encoding='UTF-8'?> \n" +
"<message>\n" +
"  Hello world!\n" +
"</message>">

diff:
? <?xml version='1.0' encoding='UTF-8'?>
  <message>
    Hello world!
  </message>

```

This is happen because `REXML::Formatters::Pretty#write_document` has a logic that depends on the last text node.

We should ignore all top-level text nodes with pretty format.

### DIFF
--- a/lib/rexml/formatters/pretty.rb
+++ b/lib/rexml/formatters/pretty.rb
@@ -111,7 +111,7 @@ module REXML
         # itself, then we don't need a carriage return... which makes this
         # logic more complex.
         node.children.each { |child|
-          next if child == node.children[-1] and child.instance_of?(Text)
+          next if child.instance_of?(Text)
           unless child == node.children[0] or child.instance_of?(Text) or
             (child == node.children[1] and !node.children[0].writethis)
             output << "\n"
--- a/test/test_document.rb
+++ b/test/test_document.rb
@@ -247,7 +247,7 @@ EOX
 
     class WriteTest < Test::Unit::TestCase
       def setup
-        @document = REXML::Document.new(<<-EOX)
+        @document = REXML::Document.new(<<-EOX.chomp)
 <?xml version="1.0" encoding="UTF-8"?>
 <message>Hello world!</message>
 EOX
@@ -257,7 +257,7 @@ EOX
         def test_output
           output = ""
           @document.write(output)
-          assert_equal(<<-EOX, output)
+          assert_equal(<<-EOX.chomp, output)
 <?xml version='1.0' encoding='UTF-8'?>
 <message>Hello world!</message>
 EOX
@@ -280,7 +280,7 @@ EOX
           indent = 2
           transitive = true
           @document.write(output, indent, transitive)
-          assert_equal(<<-EOX, output)
+          assert_equal(<<-EOX.chomp, output)
 <?xml version='1.0' encoding='UTF-8'?>
 <message
 >Hello world!</message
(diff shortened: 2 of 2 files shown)


lib/rexml/formatters/pretty.rb
```diff
@@ -111,7 +111,7 @@ def write_document( node, output )
         # itself, then we don't need a carriage return... which makes this
         # logic more complex.
         node.children.each { |child|
-          next if child == node.children[-1] and child.instance_of?(Text)
+          next if child == "\n" and child.instance_of?(Text)
```
kou, line 114: Hmm. Does this work with the following XML?

```xml
<?xml version="1.0" encoding="UTF-8"?>

<message>Hello world!</message>
```


It seems that we can ignore all top-level text nodes:

```diff
diff --git a/lib/rexml/formatters/pretty.rb b/lib/rexml/formatters/pretty.rb
index a1198b7..a838d83 100644
--- a/lib/rexml/formatters/pretty.rb
+++ b/lib/rexml/formatters/pretty.rb
@@ -111,7 +111,7 @@ module REXML
         # itself, then we don't need a carriage return... which makes this
         # logic more complex.
         node.children.each { |child|
-          next if child == node.children[-1] and child.instance_of?(Text)
+          next if child.instance_of?(Text)
           unless child == node.children[0] or child.instance_of?(Text) or
             (child == node.children[1] and !node.children[0].writethis)
             output << "\n"
```

naitoh: I think this is the right correction, too.

If `next if child.instance_of?(Text)`, then the following XML works.
```
<?xml version="1.0" encoding="UTF-8"?>

<message>Hello world!</message>
```

Fryguy: @naitoh Interestingly if you have top-level text nodes that are _not_ an `?xml` node, then you end up with the strange behavior below where it outputs a leading blank line.

```xml
XXX<message>Hello world!</message>
```

```ruby
d = REXML::Document.new("XXX<message>Hello world!</message>")
d.to_s
# => "XXX<message>Hello world!</message>"
d.write(out = '', 2)
out
# => "\n<message>\n  Hello world!\n</message>"
```

Note that I think this leading "XXX" is invalid XML anway, but is parsed successfully with REXML 3.3.2 (and older).  I wonder if you also need to fail on text nodes _before_ the root node similar to how you did it in #161 for those after the root node?

naitoh: > I wonder if you also need to fail on text nodes _before_ the root node similar to how you did it in #161 for those after the root node?

```XXX<message>Hello world!</message>```

This is invalid XML and I think should be an error.

@kou 
What do you think?

kou: Let's open a new issue and discuss this on the new issue.

naitoh: I have created https://github.com/ruby/rexml/pull/184 which fixes the above.

PR #165, opened 2024-07-10
Add tests for REXML::Text.check

Watson1978:
This patch will add missing REXML::Text.check tests.

 This is the tests for the part that is checked using a regular expression:
 https://github.com/ruby/rexml/blob/b2ec329dc1dc7635b224a6d61687c24b1e1db6fd/lib/rexml/text.rb#L155-L172

### DIFF
--- a/test/test_text_check.rb
+++ b/test/test_text_check.rb
@@ -68,21 +68,6 @@ module REXMLTests
         assert_check_failed('&#8;', '&#8;')
       end
 
-      def test_character_reference_format_without_hash
-        # U+0030 DIGIT ZERO
-        assert_check_failed('&48;', '&')
-      end
-
-      def test_character_reference_format_without_hash_and_start_0
-        # U+0030 DIGIT ZERO
-        assert_check_failed('&048;', '&')
-      end
-
-      def test_character_reference_format_without_hash_and_start_00
-        # U+0030 DIGIT ZERO
-        assert_check_failed('&0048;', '&')
-      end
-
       def test_character_reference_format_hex_0x
         # U+0041 LATIN CAPITAL LETTER A
         assert_check_failed('&#0x41;', '&#0x41;')
--- a/test/test_text_check.rb
+++ b/test/test_text_check.rb
@@ -55,6 +55,10 @@ module REXMLTests
         assert_check_failed('<;', '<')
       end
 
+      def test_lt_mix
+        assert_check_failed('ab<cd', '<')
+      end
+
       def test_entity_reference_missing_colon
         assert_check_failed('&amp', '&')
       end
@@ -62,6 +66,31 @@ module REXMLTests
       def test_character_reference_decimal_invalid_value
         # U+0008 BACKSPACE
         assert_check_failed('&#8;', '&#8;')
+      end
+
+      def test_character_reference_format_without_hash
+        # U+0030 DIGIT ZERO
+        assert_check_failed('&48;', '&')
+      end
+
+      def test_character_reference_format_without_hash_and_start_0
+        # U+0030 DIGIT ZERO
+        assert_check_failed('&048;', '&')
+      end
+
+      def test_character_reference_format_without_hash_and_start_00
+        # U+0030 DIGIT ZERO
+        assert_check_failed('&0048;', '&')
+      end
+
+      def test_character_reference_format_hex_0x
+        # U+0041 LATIN CAPITAL LETTER A
+        assert_check_failed('&#0x41;', '&#0x41;')
+      end
+
+      def test_character_reference_format_hex_00x
+        # U+0041 LATIN CAPITAL LETTER A
+        assert_check_failed('&#00x41;', '&#00x41;')
       end
 
       def test_character_reference_hex_invalid_value
--- a/test/test_text_check.rb
+++ b/test/test_text_check.rb
@@ -40,6 +40,10 @@ module REXMLTests
         # U+3044 HIRAGANA LETTER I
         assert_nothing_raised { check("&\u3042\u3044;") }
       end
+
+      def test_normal_string
+        assert_nothing_raised { check("foo") }
+      end
     end
 
     class TestInvalid < self
--- a/test/test_text_check.rb
+++ b/test/test_text_check.rb
@@ -2,49 +2,73 @@
 
 module REXMLTests
   class TextCheckTester < Test::Unit::TestCase
-    include REXML
 
-    def test_valid_pattern
-      chars = ['&A;', '&:;', '&_;', '&:A;', '&AA.bb-00;']
-      chars.each do |token|
-        text = Text.new(token, false, nil, true)
-        assert_equal(token, text.to_s)
+    def check(string)
+      REXML::Text.check(string, REXML::Text::NEEDS_A_SECOND_CHECK, nil)
+    end
+
+    class TestValid < self
+      def test_entity_name_start_char_colon
+        string = '&:;'
+        text = check(string)
+        assert_equal(string, text.to_s)
       end
 
-      character_entities = ['&amp;', '&lt;', '&gt;', '&quot;', '&apos;', '&nbsp;', '&Psi;']
-      character_entities.each do |token|
-        text = Text.new(token, false, nil, true)
-        assert_equal(token, text.to_s)
+      def test_entity_name_start_char_under_score
+        string = '&_;'
+        text = check(string)
+        assert_equal(string, text.to_s)
       end
 
-      numeric_entities = ['&#13;', '&#34;', '&#9830;', '&#0000000013;', '&#0000000162;', '&#x10FFFF;', '&#1114111;']
-      numeric_entities.each do |token|
-        text = Text.new(token, false, nil, true)
-        assert_equal(token, text.to_s)
+      def test_entity_name_char
+        string = '&A.b-0123;'
+        text = check(string)
+        assert_equal(string, text.to_s)
       end
 
-      unicode_entities = ['&#x9;', '&#xa;', '&#xD;', '&#x84;', '&#x9F;', '&#xFDEF;', '&#x10FFFF;', '&#x000000007f;']
-      unicode_entities.each do |token|
-        text = Text.new(token, false, nil, true)
-        assert_equal(token, text.to_s)
+      def test_numeric_entity_decimal
+        string = '&#0162;'
+        text = check(string)
+        assert_equal(string, text.to_s)
       end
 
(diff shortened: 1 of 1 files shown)


test/test_text_check.rb
```diff
@@ -0,0 +1,46 @@
...
+module REXMLTests
+  class TextCheckTester < Test::Unit::TestCase
+    include REXML
+
+    def test_valid_pattern
+      chars = ['&A;', '&:;', '&_;', '&:A;', '&AA.bb-00;']
+      chars.each do |token|
+        text = Text.new(token, false, nil, true)
+        assert_equal(token, text.to_s)
+      end
```
kou, line 12: Could you create a test per pattern?

```ruby
def check(string)
  REXML::Text.check(string, REXML::Text::NEEDS_A_SECOND_CHECK, nil)
end

class TestValid < self
  class TestCharacterReference < self
    def test_unknown_ascii
      check("&A;")
    end

    def test_unknown_symbol
      check("&:;")
    end
    
    def test_defined_amp
      check("&amp;")
    end

    # ...
```

test/test_text_check.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: false
+
+module REXMLTests
+  class TextCheckTester < Test::Unit::TestCase
+    include REXML
```
kou, line 5: Could you avoid `include`?

test/test_text_check.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    def test_valid_pattern
+      chars = ['&A;', '&:;', '&_;', '&:A;', '&AA.bb-00;']
```
kou, line 8: Do we need both of `&:;` and `&_;`?
Do we need `&:A;`?

test/test_text_check.rb
```diff
@@ -0,0 +1,46 @@
...
+      chars.each do |token|
+        text = Text.new(token, false, nil, true)
+        assert_equal(token, text.to_s)
+      end
+
+      character_entities = ['&amp;', '&lt;', '&gt;', '&quot;', '&apos;', '&nbsp;', '&Psi;']
```
kou, line 14: Do we need all patterns?

test/test_text_check.rb
```diff
@@ -0,0 +1,46 @@
...
+      character_entities.each do |token|
+        text = Text.new(token, false, nil, true)
+        assert_equal(token, text.to_s)
+      end
+
+      numeric_entities = ['&#13;', '&#34;', '&#9830;', '&#0000000013;', '&#0000000162;', '&#x10FFFF;', '&#1114111;']
```
kou, line 20: Can we reduce patterns?

test/test_text_check.rb
```diff
@@ -0,0 +1,46 @@
...
+      numeric_entities.each do |token|
+        text = Text.new(token, false, nil, true)
+        assert_equal(token, text.to_s)
+      end
+
+      unicode_entities = ['&#x9;', '&#xa;', '&#xD;', '&#x84;', '&#x9F;', '&#xFDEF;', '&#x10FFFF;', '&#x000000007f;']
```
kou, line 26: Can we reduce patterns?

test/test_text_check.rb
```diff
@@ -0,0 +1,46 @@
...
+      unicode_entities.each do |token|
+        text = Text.new(token, false, nil, true)
+        assert_equal(token, text.to_s)
+      end
+
+      unicodes = ["&\u00C0;", "&\uFDF0;", "&\u{10000};", "&\u00D6\u0300\u0300;"]
```
kou, line 32: Do we need all patterns?

test/test_text_check.rb
```diff
@@ -0,0 +1,46 @@
...
+      unicodes.each do |unicode|
+        text = Text.new(unicode, false, nil, true)
+        assert_equal(unicode, text.to_s)
+      end
+    end
+
+    def test_invalid_pattern
+      chars = ['<', '<;', '&amp', '&42;','&#A;', '&#8;', '&#xB;', '&#x1f;', '&#xD800;', '&#xFFFE;', '&#x110000;', '&#1114112;']
+      chars.each do |token|
+        assert_raise(RuntimeError) { Text.new(token, false, nil, true) }
```
kou, line 42: Could you also check error message?

```suggestion
        assert_raise(RuntimeError.new("XXX")) { Text.new(token, false, nil, true) }
```

test/test_text_check.rb
Watson1978: > Do we need both of &:; and &_;?
The characters that can be used individually are specified in:
https://github.com/ruby/rexml/blob/b2ec329dc1dc7635b224a6d61687c24b1e1db6fd/lib/rexml/xmltokens.rb#L26-L29

> Do we need &:A;?
Hmm, we can remove it...

test/test_text_check.rb
Watson1978: I split the test cases at https://github.com/ruby/rexml/pull/165/commits/c2b376f405f1f9b7cce08c9adb2b1527a806dfd8 

And I reduced the test patterns.

test/test_text_check.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      def test_unicode_entity
+        string = "&\u00D6\u0300\u0300;"
+        text = check(string)
+        assert_equal(string, text.to_s)
+      end
+    end
+
+    class TestInvalid < self
+      def test_lt
```
kou, line 49: Could you also add a pattern that exists `<` not at the beginning of the string?

test/test_text_check.rb
```diff
@@ -0,0 +1,75 @@
...
+        string = "<;"
+        assert_raise(RuntimeError.new("Illegal character \"<\" in raw string #{string.inspect}")) { check(string) }
+      end
+
+      def test_missing_colon
```
kou, line 54:
```suggestion
      def test_entity_reference_missing_colon
```

test/test_text_check.rb
```diff
@@ -0,0 +1,75 @@
...
+        string = "&amp"
+        assert_raise(RuntimeError.new("Illegal character \"&\" in raw string #{string.inspect}")) { check(string) }
+      end
+
+      def test_invalid_numeric_entity_decimal
```
kou, line 59: Could you add a reason why this value is invalid to test name?

```suggestion
      def test_character_reference_decimal_control_character
```

test/test_text_check.rb
```diff
@@ -0,0 +1,75 @@
...
+        string = "&#8;"
```
kou, line 60: Could you add a comment what is this character?
For example:

```ruby
# BACKSPACE
```

test/test_text_check.rb
```diff
@@ -0,0 +1,75 @@
...
+        assert_raise(RuntimeError.new("Illegal character #{string.inspect} in raw string #{string.inspect}")) { check(string) }
+      end
+
+      def test_invalid_numeric_entity_hex
+        string = "&#xD800;"
```
kou, line 65: ditto.

test/test_text_check.rb
```diff
@@ -0,0 +1,75 @@
...
+        assert_raise(RuntimeError.new("Illegal character #{string.inspect} in raw string #{string.inspect}")) { check(string) }
+      end
+
+      def test_invalid_unicode
+        string = "&\u00BF;"
```
kou, line 70: ditto.

test/test_text_check.rb
```diff
@@ -0,0 +1,75 @@
...
+        string = "&amp"
+        assert_raise(RuntimeError.new("Illegal character \"&\" in raw string #{string.inspect}")) { check(string) }
+      end
+
+      def test_invalid_numeric_entity_decimal
+        string = "&#8;"
+        assert_raise(RuntimeError.new("Illegal character #{string.inspect} in raw string #{string.inspect}")) { check(string) }
+      end
+
+      def test_invalid_numeric_entity_hex
```
kou, line 64: Could you also add `&#0x...;` and `&#00x...;` patterns?

test/test_text_check.rb
kou, line 60: Could you also add `&0...;` and `&00...;` patterns?

test/test_text_check.rb
```diff
@@ -0,0 +1,107 @@
...
+
+module REXMLTests
+  class TextCheckTester < Test::Unit::TestCase
+
+    def check(string)
+      REXML::Text.check(string, REXML::Text::NEEDS_A_SECOND_CHECK, nil)
+    end
+
+    def assert_check(string)
+      assert_nothing_raised { REXML::Text.check(string, REXML::Text::NEEDS_A_SECOND_CHECK, nil) }
```
kou, line 11:
```suggestion
      assert_nothing_raised { check(string) }
```

test/test_text_check.rb
```diff
@@ -0,0 +1,107 @@
...
+
+      def test_character_reference_format_without_hash_and_start_0
+        # U+0030 DIGIT ZERO
+        assert_check_failed('&048;', '&')
+      end
+
+      def test_character_reference_format_without_hash_and_start_00
+        # U+0030 DIGIT ZERO
+        assert_check_failed('&0048;', '&')
+      end
```
kou, line 84: We don't need no `#` cases because it's entity reference case not character reference case.

test/test_text_check.rb
```diff
@@ -0,0 +1,107 @@
...
+        # U+0041 LATIN CAPITAL LETTER A
+        assert_check_failed('&#0x41;', '&#0x41;')
+      end
+
+      def test_character_reference_format_hex_00x
+        # U+0041 LATIN CAPITAL LETTER A
+        assert_check_failed('&#00x41;', '&#00x41;')
+      end
+
+      def test_character_reference_hex_invalid_value
```
kou, line 96: Could you use the reason instead of "invalid_value"? "test_..._surrogate_block"?
https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-Char

> 	/* any Unicode character, excluding the surrogate blocks, FFFE, and FFFF. */

test/test_text_check.rb
```diff
@@ -0,0 +1,107 @@
...
+        # U+0D800 SURROGATE PAIR
+        assert_check_failed('&#xD800;', '&#xD800;')
+      end
+
+      def test_entity_name_non_ascii_invalid_value
```
kou, line 101: ditto.

Watson1978: It is dealing with is not a surrogate, but a symbol that is not allowed in XML.

kou:
Thanks.

PR #166, opened 2024-07-11
Fix test for Text.check

Watson1978:
This patch will fix incorrect string in a case where unicode characters.
Because of the use of single quotes, it was simply an ASCII string.


kou:
Could you always use `"` to avoid similar mistakes?

Watson1978:
Replaced all `'` to `"` in test_text_check.rb

kou:
Thanks.

PR #169, opened 2024-07-16
Fix ReDoS caused by very large character references using repeated 0s

Watson1978:
This patch will fix the ReDoS that is caused by large string of 0s on a character reference (like `&#00000000...`).

This is occurred in Ruby 3.1 or earlier.

### DIFF
--- a/test/parse/test_character_reference.rb
+++ b/test/parse/test_character_reference.rb
@@ -7,7 +7,7 @@ module REXMLTests
   class TestParseCharacterReference < Test::Unit::TestCase
     include Test::Unit::CoreAssertions
 
-    def test_gt_linear_performance_character_reference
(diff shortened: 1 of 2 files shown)


test/parse/test_character_reference.rb
```diff
@@ -0,0 +1,17 @@
+require "test/unit"
+require "core_assertions"
+
+require "rexml/document"
+
+module REXMLTests
+  class TestParseCharacterReference < Test::Unit::TestCase
+    include Test::Unit::CoreAssertions
+
+    def test_gt_linear_performance_character_reference
```
kou, line 10:
```suggestion
    def test_gt_linear_performance_many_preceding_zeros
```

kou:
Thanks.

PR #170, opened 2024-07-16
Fix performance issue caused by using repeated `>` characters inside `<?xml`

Watson1978:


test/parse/test_processing_instruction.rb
```diff
@@ -69,5 +73,12 @@ def test_after_root
 
       assert_equal("abc", events[:processing_instruction])
     end
+
+    def test_gt_linear_performance_introduction
```
kou, line 77:
```suggestion
    def test_gt_linear_performance
```

kou:
Thanks.

PR #171, opened 2024-07-16
Fix performance issue caused by using repeated `>` characters inside comments

Watson1978:


test/parse/test_comment.rb
```diff
@@ -117,5 +121,12 @@ def test_after_root
 
       assert_equal(" ok comment ", events[:comment])
     end
+
+    def test_gt_linear_performance_comment
```
kou, line 125:
```suggestion
    def test_gt_linear_performance
```

kou:
Thanks.

PR #172, opened 2024-07-16
Fix performance issue caused by using repeated `>` characters inside `CDATA [ PAYLOAD ]`

Watson1978:

### DIFF
--- a/test/parse/test_cdata.rb
+++ b/test/parse/test_cdata.rb
@@ -1,0 +1,17 @@
+require "test/unit"
+require "core_assertions"
+
+require "rexml/document"
+
+module REXMLTests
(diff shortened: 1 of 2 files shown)


test/parse/test_cdata.rb
```diff
@@ -0,0 +1,17 @@
+require "test/unit"
+require "core_assertions"
+
+require "rexml/document"
+
+module REXMLTests
+  class TestParseCData < Test::Unit::TestCase
+    include Test::Unit::CoreAssertions
+
+    def test_gt_linear_performance_cdata
```
kou, line 10:
```suggestion
    def test_gt_linear_performance
```

kou:
Thanks.

PR #174, opened 2024-07-16
Fix performance issue caused by using repeated `>` characters inside `<!DOCTYPE root [<!-- PAYLOAD -->]>`

Watson1978:


kou:
Thanks.

PR #175, opened 2024-07-16
Fix performance issue caused by using repeated `>` characters inside `<!DOCTYPE name [<!ENTITY>]>`

Watson1978:


test/parse/test_entity_declaration.rb
```diff
@@ -32,5 +32,12 @@ def test_empty
  <!ENTITY>  ]> <r/>
       DETAIL
     end
+
+    def test_gt_linear_performance_entity
```
kou, line 36:
```suggestion
    def test_gt_linear_performance
```

PR #176, opened 2024-07-16
Fix ReDoS by using repeated space characters inside `<!DOCTYPE name [<!ATTLIST>]>`

Watson1978:
Fix performance by removing unnecessary spaces.

This is occurred in Ruby 3.1 or earlier.

### DIFF
--- a/test/parse/test_attlist.rb
+++ b/test/parse/test_attlist.rb
@@ -7,7 +7,7 @@ module REXMLTests
   class TestParseAttlist < Test::Unit::TestCase
     include Test::Unit::CoreAssertions
 
-    def test_gt_linear_performance_attlist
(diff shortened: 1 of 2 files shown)


test/parse/test_attlist.rb
```diff
@@ -0,0 +1,17 @@
+require "test/unit"
+require "core_assertions"
+
+require "rexml/document"
+
+module REXMLTests
+  class TestParseAttlist < Test::Unit::TestCase
+    include Test::Unit::CoreAssertions
+
+    def test_gt_linear_performance_attlist
```
kou, line 10:
```suggestion
    def test_gt_linear_performance
```

kou:
Thanks.

PR #177, opened 2024-07-16
Fix performance issue caused by using repeated `>` characters inside `<xml><!-- --></xml>`

Watson1978:

### DIFF
--- a/test/parse/test_comment.rb
+++ b/test/parse/test_comment.rb
@@ -129,7 +129,7 @@ module REXMLTests
       end
     end
 
-    def test_gt_linear_performance_comment_in_element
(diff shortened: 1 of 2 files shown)


test/parse/test_comment.rb
```diff
@@ -128,5 +128,12 @@ def test_gt_linear_performance
         REXML::Document.new('<!-- ' + ">" * n + ' -->')
       end
     end
+
+    def test_gt_linear_performance_comment_in_element
```
kou, line 132:
```suggestion
    def test_gt_linear_performance_in_element
```

kou:
Thanks.

PR #182, opened 2024-07-19
Fix method scope in test in order to invoke the tests properly and fix exception message

Watson1978:
This PR includes following two fixes.

1. The `test_empty` and `test_linear_performance_gt` were defined as private method. Seems that test-unit runner does not invoke private methods even if the methods have `test_` prefix.
2. When parse malformed entity declaration, the exception might have the message about `NoMethodError`. The proper exception message will be contained by this fix.

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -318,11 +318,11 @@ module REXML
               raise REXML::ParseException.new( "Bad ELEMENT declaration!", @source ) if md.nil?
               return [ :elementdecl, "<!ELEMENT" + md[1] ]
             elsif @source.match("ENTITY", true)
-              scanner = @source.match(Private::ENTITYDECL_PATTERN, true, term: Private::ENTITY_TERM)
-              unless scanner
+              match_data = @source.match(Private::ENTITYDECL_PATTERN, true, term: Private::ENTITY_TERM)
+              unless match_data
                 raise REXML::ParseException.new("Malformed entity declaration", @source)
               end
-              match = [:entitydecl, *scanner.captures.compact]
+              match = [:entitydecl, *match_data.captures.compact]
               ref = false
               if match[1] == '%'
                 ref = true
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -318,7 +318,11 @@ module REXML
               raise REXML::ParseException.new( "Bad ELEMENT declaration!", @source ) if md.nil?
               return [ :elementdecl, "<!ELEMENT" + md[1] ]
             elsif @source.match("ENTITY", true)
-              match = [:entitydecl, *@source.match(Private::ENTITYDECL_PATTERN, true, term: Private::ENTITY_TERM).captures.compact]
+              scanner = @source.match(Private::ENTITYDECL_PATTERN, true, term: Private::ENTITY_TERM)
+              unless scanner
+                raise REXML::ParseException.new("Malformed entity declaration", @source)
+              end
+              match = [:entitydecl, *scanner.captures.compact]
               ref = false
               if match[1] == '%'
                 ref = true
--- a/test/parse/test_entity_declaration.rb
+++ b/test/parse/test_entity_declaration.rb
@@ -24,11 +24,18 @@ module REXMLTests
 
     public
     def test_empty
-      assert_raise(REXML::ParseException) do
+      exception = assert_raise(REXML::ParseException) do
         parse(<<-INTERNAL_SUBSET)
 <!ENTITY>
         INTERNAL_SUBSET
       end
+      assert_equal(<<-DETAIL.chomp, exception.to_s)
+Malformed entity declaration
+Line: 5
+Position: 70
+Last 80 unconsumed characters:
+>  ]> <r/> 
+      DETAIL
     end
 
     def test_linear_performance_gt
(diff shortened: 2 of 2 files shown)


test/parse/test_entity_declaration.rb
```diff
@@ -18,25 +22,22 @@ def parse(internal_subset)
...
 <!ENTITY>
         INTERNAL_SUBSET
       end
-      assert_equal(<<-DETAIL.chomp, exception.to_s)
-Malformed notation declaration: name is missing
-Line: 5
-Position: 72
-Last 80 unconsumed characters:
- <!ENTITY>  ]> <r/>
-      DETAIL
```
Watson1978, line 33: This test raises following exception:

```
<"#<NoMethodError: undefined method `captures' for nil>\n" +
"/home/watson/prj/rexml/lib/rexml/parsers/baseparser.rb:321:in `pull_event'\n" +
"/home/watson/prj/rexml/lib/rexml/parsers/baseparser.rb:225:in `pull'\n" +
"/home/watson/prj/rexml/lib/rexml/parsers/treeparser.rb:22:in `parse'\n" +
"/home/watson/prj/rexml/lib/rexml/document.rb:448:in `build'\n" +
```

I removed because I didn't think the need to check `NoMethodError` message.

kou: We should raise `ParseException`.
Could you fix the implementation in this PR?

test/parse/test_entity_declaration.rb
```diff
@@ -18,25 +22,22 @@ def parse(internal_subset)
...
     end
 
     def test_linear_performance_gt
       seq = [10000, 50000, 100000, 150000, 200000]
       assert_linear_performance(seq, rehearsal: 10) do |n|
-        REXML::Document.new('<!DOCTYPE rubynet [<!ENTITY rbconfig.ruby_version "' + '>' * n + '">')
+        begin
+          REXML::Document.new('<!DOCTYPE rubynet [<!ENTITY rbconfig.ruby_version "' + '>' * n + '">')
```
kou, line 38: Can we use a valid XML to reproduce this?

test/parse/test_entity_declaration.rb
Watson1978: Ah, I was wrong above comment.  sorry.

It raises `ParseException`, but it has  `#<NoMethodError: undefined method `captures' for nil>` exception message.

```
#<REXML::ParseException:"#<NoMethodError: undefined method `captures' for nil>
```

kou: OK. Then, could you fix the `NoMethodError` error message in this PR?

Watson1978: I fixed with https://github.com/ruby/rexml/pull/182/commits/93ec28f68b6c04f6c31bc8e117b586ef1a469331

test/parse/test_entity_declaration.rb
Watson1978: I fixed with https://github.com/ruby/rexml/pull/182/commits/36ba79eb0fd01c68b37509772d922a80a0f9d530

lib/rexml/parsers/baseparser.rb
```diff
@@ -318,7 +318,11 @@ def pull_event
               raise REXML::ParseException.new( "Bad ELEMENT declaration!", @source ) if md.nil?
               return [ :elementdecl, "<!ELEMENT" + md[1] ]
             elsif @source.match("ENTITY", true)
-              match = [:entitydecl, *@source.match(Private::ENTITYDECL_PATTERN, true, term: Private::ENTITY_TERM).captures.compact]
+              scanner = @source.match(Private::ENTITYDECL_PATTERN, true, term: Private::ENTITY_TERM)
```
kou, line 321: Could you use `match_data` not `scanner`?
We want to use `MatchData` instead of `StringScanner` because we don't want to use `StringScanner` in `REXML::Source` directly. But `REXML::Source#match` returns `StringScanner` directly for performance reason.
So we should assume that the return value of `REXML::Source#match` is a `MatchData` and we should only use `MatchData` compatible API provided by `StringScanner`.

test/parse/test_entity_declaration.rb
```diff
@@ -18,25 +22,26 @@ def parse(internal_subset)
...
       end
       assert_equal(<<-DETAIL.chomp, exception.to_s)
-Malformed notation declaration: name is missing
+Malformed entity declaration
 Line: 5
-Position: 72
+Position: 70
 Last 80 unconsumed characters:
- <!ENTITY>  ]> <r/>
+>  ]> <r/>
```
otegami, line 37: [nits] There is a unnecessary white space.
```suggestion
>  ]> <r/>
```

lib/rexml/parsers/baseparser.rb
Watson1978: I see, so that's the intention.
https://github.com/ruby/rexml/pull/182/commits/563bb5737d0e8420d70e78d25f330327bfde5a9a

test/parse/test_entity_declaration.rb
Watson1978: Unfortunately, the exception message contains trailing space. 
So I cannot remove it.

otegami: Ahh I see. I'm so sorry mm

kou:
Could you update the description that this PR also includes `NoMethodError` fix before we merge this?

Watson1978:
> Could you update the description that this PR also includes `NoMethodError` fix before we merge this?

@kou OK, I updated the description. Thanks.

kou:
Thanks.

PR #186, opened 2024-07-30
Fix source.match performance without specifying term string

tompng:
Performance problem of `source.match(regexp)` was recently fixed by specifying terminator string. However, I think maintaining appropriate terminator string for a regexp is hard.
I propose solving this performance issue by increasing bytes to read in each iteration.

### DIFF
--- a/lib/rexml/source.rb
+++ b/lib/rexml/source.rb
@@ -247,9 +247,6 @@ module REXML
       read if @scanner.eos? && @source
     end
 
-    # Note: When specifying a string for 'pattern', it must not include '>' except in the following formats:
-    # - ">"
-    # - "XXX>" (X is any string excluding '>')
     def match( pattern, cons=false )
       # To avoid performance issue, we need to increase bytes to read per scan
       min_bytes = 1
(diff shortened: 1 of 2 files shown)


lib/rexml/source.rb
```diff
@@ -240,7 +250,9 @@ def ensure_buffer
     # Note: When specifying a string for 'pattern', it must not include '>' except in the following formats:
     # - ">"
     # - "XXX>" (X is any string excluding '>')
```
kou, line 252: Could you remove this outdated comment?

lib/rexml/source.rb
```diff
@@ -240,7 +250,9 @@ def ensure_buffer
...
-    def match( pattern, cons=false, term: nil )
+    def match( pattern, cons=false )
```
olleolleolle, line 253: Style. Very minor: the spaces around the argument list does not match the other methods in the file.

kou: Yes. We should use a consistent coding style in the same code base. But it's out-of-scope of this pull request. We should work on it separately.

lib/rexml/source.rb
tompng: Removed :+1:

kou:
Thanks.

PR #191, opened 2024-08-01
Fix to not allow parameter entity references at internal subsets

naitoh:
## Why?
In the internal subset of DTD, references to parameter entities are not allowed within markup declarations.

See: https://www.w3.org/TR/xml/#wfc-PEinInternalSubset

> Well-formedness constraint: PEs in Internal Subset

> In the internal DTD subset, parameter-entity references MUST NOT occur within markup declarations; they may occur where markup declarations can occur. (This does not apply to references that occur in external parameter entities or to the external subset.)

### DIFF
--- a/lib/rexml/entity.rb
+++ b/lib/rexml/entity.rb
@@ -12,6 +12,7 @@ module REXML
     EXTERNALID = "(?:(?:(SYSTEM)\\s+#{SYSTEMLITERAL})|(?:(PUBLIC)\\s+#{PUBIDLITERAL}\\s+#{SYSTEMLITERAL}))"
     NDATADECL = "\\s+NDATA\\s+#{NAME}"
     PEREFERENCE = "%#{NAME};"
+    PEREFERENCE_RE = /#{PEREFERENCE}/um
     ENTITYVALUE = %Q{((?:"(?:[^%&"]|#{PEREFERENCE}|#{REFERENCE})*")|(?:'([^%&']|#{PEREFERENCE}|#{REFERENCE})*'))}
     PEDEF = "(?:#{ENTITYVALUE}|#{EXTERNALID})"
     ENTITYDEF = "(?:#{ENTITYVALUE}|(?:#{EXTERNALID}(#{NDATADECL})?))"
@@ -19,7 +20,7 @@ module REXML
     GEDECL = "<!ENTITY\\s+#{NAME}\\s+#{ENTITYDEF}\\s*>"
     ENTITYDECL = /\s*(?:#{GEDECL})|(?:#{PEDECL})/um
 
-    attr_reader :name, :external, :ref, :ndata, :pubid
+    attr_reader :name, :external, :ref, :ndata, :pubid, :value
 
     # Create a new entity.  Simple entities can be constructed by passing a
     # name, value to the constructor; this creates a generic, plain entity
@@ -68,14 +69,11 @@ module REXML
     end
 
     # Evaluates to the unnormalized value of this entity; that is, replacing
-    # all entities -- both %ent; and &ent; entities.  This differs from
-    # +value()+ in that +value+ only replaces %ent; entities.
+    # &ent; entities.
     def unnormalized
       document.record_entity_expansion unless document.nil?
-      v = value()
-      return nil if v.nil?
-      @unnormalized = Text::unnormalize(v, parent)
-      @unnormalized
+      return nil if @value.nil?
+      @unnormalized = Text::unnormalize(@value, parent)
     end
 
     #once :unnormalized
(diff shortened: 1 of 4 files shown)


kou (commented):
Wow!

Can we also check this in `BaseParser`?
If we can do it, can we move the validation to `#write`?

Do we need to update `#unnormalized` and `#normalized` too?
https://github.com/ruby/rexml/blob/e3f747fb4fe30f5c890a4bea5b12dd72f595e6b3/lib/rexml/entity.rb#L70-L79
https://github.com/ruby/rexml/blob/e3f747fb4fe30f5c890a4bea5b12dd72f595e6b3/lib/rexml/entity.rb#L83-L87

lib/rexml/entity.rb
```diff
@@ -123,43 +123,14 @@ def to_s
...
-        matches.each do |entity_reference|
-          entity_value = @parent.entity( entity_reference[0] )
-          if sum + entity_value.bytesize > Security.entity_expansion_text_limit
-            raise "entity expansion has grown too large"
-          else
-            sum += entity_value.bytesize
-          end
-          rv.gsub!( /%#{entity_reference.join};/um, entity_value )
-        end
+      if @value.match?(PEREFERENCE_RE)
```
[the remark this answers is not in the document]
naitoh: OK, I see.

naitoh:
> Wow!
> 
> Can we also check this in `BaseParser`?

I changed BaseParser to do this check. 

> If we can do it, can we move the validation to `#write`?

I added validation to Entity.new.

> Do we need to update `#unnormalized` and `#normalized` too?

Updated `#unnormalized`.
No need to update `#normalized`.

lib/rexml/entity.rb
```diff
@@ -59,6 +60,10 @@ def initialize stream, value=nil, parent=nil, reference=false
         @name = stream
         @value = value
       end
+
+      if PEREFERENCE_RE.match?(@value)
+        raise "Parameter Entity References forbidden in internal subset: #{@value}"
```
kou, line 65: * Could you use `ArgumentError` instead of `RuntimeError`?
* Could you avoid needless capital case?

```suggestion
        raise ArgumentError, "Parameter entity references forbidden in internal subset: #{@value}"
```

BTW, do we need a validation here? This method document says the following:

> +WARNING+: There is no validation of entity state except when the entity
> is read from a stream.  If you start poking around with the accessors,
> you can easily create a non-conformant Entity.

lib/rexml/parsers/baseparser.rb
```diff
@@ -125,6 +125,7 @@ class BaseParser
       }
 
       module Private
+        PEREFERENCE_RE = /#{PEREFERENCE}/um
```
kou, line 128:
```suggestion
        PEREFERENCE_PATTERN = /#{PEREFERENCE}/um
```

lib/rexml/parsers/baseparser.rb
```diff
@@ -334,6 +335,8 @@ def pull_event
                 match[4] = match[4][1..-2] # HREF
                 match.delete_at(5) if match.size > 5 # Chop out NDATA decl
                 # match is [ :entity, name, PUBLIC, pubid, href(, ndata)? ]
+              elsif Private::PEREFERENCE_RE.match?(match[2])
+                raise REXML::ParseException.new("Parameter Entity References forbidden in internal subset: #{match[2]}", @source)
```
kou, line 339:
```suggestion
                raise REXML::ParseException.new("Parameter entity references forbidden in internal subset: #{match[2]}", @source)
```

test/test_entity.rb
```diff
@@ -104,6 +102,62 @@ def test_replace_entities
       assert_equal source, out
     end
 
+    def test_entity_with_reference
```
kou, line 105: Can we improve test name? This class is `EntityTester`. So `_entity` is redundant.

`test_readers_reference`?

test/test_entity.rb
```diff
@@ -104,6 +102,62 @@ def test_replace_entities
...
+      entity = REXML::Entity.new( [:entitydecl, "a", "B", "%"])
+      assert_equal('<!ENTITY % a "B">', entity.to_s)
+      assert_equal("a", entity.name)
+      assert_equal("B", entity.value)
+      assert_equal("B", entity.normalized)
+      assert_equal("B", entity.unnormalized)
```
kou, line 111: Could you check by one `assert_equal` for easy to debug?
FYI: https://www.clear-code.com/blog/2011/2/28.html

```suggestion
      assert_equal([
                     '<!ENTITY % a "B">',
                     "a",
                     "B",
                     "B",
                     "B",
                   ],
                   [
                     entity.to_s,
                     entity.name,
                     entity.value,
                     entity.normalized,
                     entity.unnormalized,
                   ])
```

lib/rexml/entity.rb
naitoh: @kou 
OK.
I see.
We prefer not to add this validation in `.new`.

Similarly, I don't think this validation should be added to `#write`.
What do you think?

kou: I agree with you.

kou:
Thanks.

PR #197, opened 2024-08-12
Add unit test for ReDoS vulnerability (CVE-2024-41123)

ch4n3-yoon:
Added a unit test (`test_linear_performance_tab_gt`) to ensure the performance remains linear when processing specially crafted inputs that could previously exploit the ReDoS vulnerability.


kou:
I think that the existing `test_linear_performance_tab` covers this case.
Do we need to this too?

kou:
We don't need this.

PR #199, opened 2024-08-14
Fix RuntimeError in `REXML::Parsers::BaseParser` for valid feeds

### LINKED ISSUE #198: RuntimeError "entity expansion has grown too large" with default entities in `REXML::Parsers::BaseParser` after 3.3.3
### Why?

We are using the [ruby/rss](https://github.com/ruby/rss) gem to parse XML feeds and we get RuntimeErrors “entity expansion has grown too large” since [rexml 3.3.3](https://github.com/ruby/rexml/releases/tag/v3.3.3) even though there is only [default entities](https://github.com/ruby/rexml/blob/e14847cee53d26eb162ad786ba12e3cd7a86fce0/lib/rexml/parsers/baseparser.rb#L120-L125) in the feeds.

### How to reproduce

Code:

```ruby
# rexml/refactor_entity_example.rb

require "rexml/parsers/baseparser"

valid_feed = "&lt;p&gt;#{'A' * 10_240}&lt;/p&gt;"

base_parser = REXML::Parsers::BaseParser.new("")
base_parser.unnormalize(valid_feed) # => "<p>" + "A" * 10_240 + "</p>"
```

Error:

```shell
❯ bundle exec ruby refactor_entity_example.rb
/Users/vikiv/repos/rexml/lib/rexml/parsers/baseparser.rb:558:in `block in unnormalize': entity expansion has grown too large (RuntimeError)
	from /Users/vikiv/repos/rexml/lib/rexml/parsers/baseparser.rb:551:in `each'
	from /Users/vikiv/repos/rexml/lib/rexml/parsers/baseparser.rb:551:in `unnormalize'
	from refactor_entity_example.rb:8:in `<main>'
```

### Suggestion  

I suggest we change `#entity`/`#unnormalize` to not raise the RuntimeError “entity expansion has grown too large” when the feed consists of only default entities. 

```diff
diff --git a/lib/rexml/parsers/baseparser.rb b/lib/rexml/parsers/baseparser.rb
index 342f948..d560d3a 100644
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -505,15 +505,13 @@ module REXML
       private :pull_event

       def entity( reference, entities )
-        value = nil
         value = entities[ reference ] if entities
         if value
           record_entity_expansion
-        else
-          value = DEFAULT_ENTITIES[ reference ]
-          value = value[2] if value
+          return unnormalize( value, entities ) if value
         end
-        unnormalize( value, entities ) if value
+
+        nil
       end

       # Escapes all possible entities
```

vikiv480:
* Change `#entity` to not match against default entities

After this change, the following example will not raise a RuntimeError:

```ruby
# rexml/refactor_entity_example.rb

require "rexml/parsers/baseparser"

valid_feed = "&lt;p&gt;#{'A' * 10_240}&lt;/p&gt;"

base_parser = REXML::Parsers::BaseParser.new("")
base_parser.unnormalize(valid_feed) # => "<p>" + "A" * 10_240 + "</p>"
```

Default entities now gets substituted by this block instead

https://github.com/ruby/rexml/blob/e14847cee53d26eb162ad786ba12e3cd7a86fce0/lib/rexml/parsers/baseparser.rb#L560-L563

Close #198

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -505,14 +505,13 @@ module REXML
       private :pull_event
 
       def entity( reference, entities )
-        value = nil
-        value = entities[ reference ] if entities
-        if value
-          record_entity_expansion
-          return unnormalize( value, entities )
-        end
+        return unless entities
 
-        nil
+        value = entities[ reference ]
+        return if value.nil?
+
+        record_entity_expansion
+        unnormalize( value, entities )
       end
 
       # Escapes all possible entities
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -505,15 +505,13 @@ module REXML
       private :pull_event
 
       def entity( reference, entities )
-        value = nil
         value = entities[ reference ] if entities
         if value
           record_entity_expansion
-        else
-          value = DEFAULT_ENTITIES[ reference ]
-          value = value[2] if value
+          return unnormalize( value, entities ) if value
         end
-        unnormalize( value, entities ) if value
+
+        nil
       end
 
       # Escapes all possible entities
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -505,10 +505,11 @@ module REXML
       private :pull_event
 
       def entity( reference, entities )
+        value = nil
         value = entities[ reference ] if entities
         if value
           record_entity_expansion
-          return unnormalize( value, entities ) if value
+          return unnormalize( value, entities )
         end
 
         nil
--- a/test/test_stream.rb
+++ b/test/test_stream.rb
@@ -102,7 +102,7 @@ module REXMLTests
 
     def test_with_only_default_entities
       member_value = "&lt;p&gt;#{'A' * @default_entity_expansion_text_limit}&lt;/p&gt;"
-      source = StringIO.new(<<-XML)
+      source = <<-XML
 <?xml version="1.0" encoding="UTF-8"?>
 <member>
 #{member_value}
--- a/test/test_stream.rb
+++ b/test/test_stream.rb
@@ -89,6 +89,44 @@ module REXMLTests
     end
   end
 
+  class EntityExpansionLimitTest < Test::Unit::TestCase
+    def setup
+      @default_entity_expansion_limit = REXML::Security.entity_expansion_limit
+      @default_entity_expansion_text_limit = REXML::Security.entity_expansion_text_limit
+    end
+
+    def teardown
+      REXML::Security.entity_expansion_limit = @default_entity_expansion_limit
+      REXML::Security.entity_expansion_text_limit = @default_entity_expansion_text_limit
+    end
+
+    def test_with_only_default_entities
+      member_value = "&lt;p&gt;#{'A' * @default_entity_expansion_text_limit}&lt;/p&gt;"
+      source = StringIO.new(<<-XML)
+<?xml version="1.0" encoding="UTF-8"?>
+<member>
+#{member_value}
+</member>
+      XML
+
+      listener = MyListener.new
+      class << listener
+        attr_accessor :text_value
+        def text(text)
+          @text_value << text
+        end
+      end
+      listener.text_value = ""
+      REXML::Document.parse_stream(source, listener)
+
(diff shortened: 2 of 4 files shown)


lib/rexml/parsers/baseparser.rb
```diff
@@ -505,15 +505,13 @@ def pull_event
       private :pull_event
 
       def entity( reference, entities )
-        value = nil
```
kou, line 508: Could you revert this change for readability?
It shows that the default `value` is `nil` explicitly.

lib/rexml/parsers/baseparser.rb
```diff
@@ -505,15 +505,13 @@ def pull_event
...
         value = entities[ reference ] if entities
         if value
           record_entity_expansion
-        else
-          value = DEFAULT_ENTITIES[ reference ]
-          value = value[2] if value
+          return unnormalize( value, entities ) if value
```
kou, line 511: Do we need this `if value`?

kou (commented):
Could you add a test for this case?

kou:
@naitoh Could you review this?

lib/rexml/parsers/baseparser.rb
vikiv480: Okay, I reverted the change in [b5da207](https://github.com/ruby/rexml/pull/199/commits/b5da2078bc7e7c6331ed3f7d8369a052d172c5fe)

lib/rexml/parsers/baseparser.rb
vikiv480: No, you are right, it's unnecessary. Removed it in [b5da207](https://github.com/ruby/rexml/pull/199/commits/b5da2078bc7e7c6331ed3f7d8369a052d172c5fe)

vikiv480:
>Could you add a test for this case?

I've added some tests in [6555c27](https://github.com/ruby/rexml/pull/199/commits/6555c279faf61b83bb5af204051f05bf1684ec0a), let me know what you think.

naitoh:
> Could you review this?

## before (This PR)
```ruby
      def entity( reference, entities )
        value = nil
        value = entities[ reference ] if entities
        if value
          record_entity_expansion
          return unnormalize( value, entities )
        end

        nil
      end
```

- If `entities == nil`, the return value is nil because value is always nil. 
- If `value == nil`, the return value is nil, so an early return is easier to understand.

## after
```ruby
      def entity( reference, entities )
        return unless entities

        value = entities[ reference ]
        return if value.nil?

        record_entity_expansion
        unnormalize( value, entities )
      end
```

Thanks.

vikiv480:
@naitoh Thanks for the suggestion, I like the change 👍 60632a5

test/test_stream.rb
```diff
@@ -89,6 +89,44 @@ def entity(content)
...
+    end
+
+    def teardown
+      REXML::Security.entity_expansion_limit = @default_entity_expansion_limit
+      REXML::Security.entity_expansion_text_limit = @default_entity_expansion_text_limit
+    end
+
+    def test_with_only_default_entities
+      member_value = "&lt;p&gt;#{'A' * @default_entity_expansion_text_limit}&lt;/p&gt;"
+      source = StringIO.new(<<-XML)
```
naitoh, line 105:
```suggestion
      source = <<-XML
```
I think StringIO.new is unnecessary.

naitoh (commented):
Additional review.

test/test_stream.rb
vikiv480: Fixed in [b7e9378](https://github.com/ruby/rexml/pull/199/commits/b7e937898847d4ab938ade808f57e47a1d08975e)

naitoh:
I have checked this PR.
I think this PR is good.
Thanks!

test/test_stream.rb
```diff
@@ -89,6 +89,44 @@ def entity(content)
...
+      REXML::Document.parse_stream(source, listener)
+
+      expected_value = "<p>#{'A' * @default_entity_expansion_text_limit}</p>"
+      assert_equal(expected_value, listener.text_value.strip)
+      assert do
+        listener.text_value.bytesize > @default_entity_expansion_text_limit
+      end
+    end
+  end
+
```
kou, line 129:
```suggestion
```

kou:
Thanks.

PR #204, opened 2024-09-04
Remove strscan dependency declaration from gemspec

### LINKED ISSUE #140: REXML is no longer pure ruby
This is a deliberate duplicate of #131, which appears to be mired in some unrelated discussion.

We use REXML because it is pure ruby and thus can be installed in environments that are limited in their native compilation support.

By requiring strscan, REXML no longer is pure Ruby.

REXML now is no longer useful for these environments.

I have pinned to <=3.2.6, but of course I'd like to have security fixes going forward.

Bo98:
See discussion in #140. `strscan` is a part of the Ruby standard library in all versions of Ruby supported by REXML.

Fixes #140.

### DIFF
--- a/rexml.gemspec
+++ b/rexml.gemspec
@@ -58,6 +58,4 @@ Gem::Specification.new do |spec|
   spec.extra_rdoc_files = rdoc_files
 
   spec.required_ruby_version = '>= 2.5.0'
-
-  spec.add_runtime_dependency("strscan")
 end


kou:
Thanks.

The Document CI failure is unrelated.

kou:
We need to fix the CI failure before we release a new version.
Could you try it?

Bo98:
Unfortunately, it's a Ruby bug it seems: https://bugs.ruby-lang.org/issues/20713.

We can workaround it though I think: #205

kou:
Thanks for taking a look at this.
I thought that nobody is working on this. So I've done by myself...

kou:
Released.

PR #205, opened 2024-09-04
Rakefile: workaround Ruby 3.3.5 bug to fix document CI

Bo98:
The initial error was a bug here on the rexml side in handling kwargs but after fixing that it revealed that the trigger for what was calling `warn` was https://bugs.ruby-lang.org/issues/20713.

We can workaround it by requiring `json` before hooking into `warn`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -14,6 +14,7 @@ task :default => :test
 namespace :warning do
   desc "Treat warning as error"
   task :error do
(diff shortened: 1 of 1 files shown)


kou:
Thanks. But sorry... I already pushed another workaround: https://github.com/ruby/rexml/commit/9294410f6eb90834a69a3fa363de61f5a3f6a927

Issue #212, opened 2024-10-08
Does REXML support `UTF-32`?


naitoh:
I don't think REXML have a test case for `UTF-32`.

Also, I get the following error

* test_utf.rb 
```
require 'rexml/document'

xml = File.open("./utf-32.xml")
REXML::Document.new(xml)
```

* utf-32.xml
```
<?xml version="1.0" encoding="UTF-32"?>
<message>Hello world!</message>
```
- [utf-32.xml.txt](https://github.com/user-attachments/files/17300209/utf-32.xml.txt) (Renamed file from `utf-32.xml`. )
```
$ hexdump utf-32.xml 
0000000 feff 0000 003c 0000 003f 0000 0078 0000
0000010 006d 0000 006c 0000 0020 0000 0076 0000
0000020 0065 0000 0072 0000 0073 0000 0069 0000
0000030 006f 0000 006e 0000 003d 0000 0022 0000
0000040 0031 0000 002e 0000 0030 0000 0022 0000
0000050 0020 0000 0065 0000 006e 0000 0063 0000
0000060 006f 0000 0064 0000 0069 0000 006e 0000
0000070 0067 0000 003d 0000 0022 0000 0055 0000
0000080 0054 0000 0046 0000 002d 0000 0033 0000
0000090 0032 0000 0022 0000 003f 0000 003e 0000
00000a0 000a 0000 003c 0000 006d 0000 0065 0000
00000b0 0073 0000 0073 0000 0061 0000 0067 0000
00000c0 0065 0000 003e 0000 0048 0000 0065 0000
00000d0 006c 0000 006c 0000 006f 0000 0020 0000
00000e0 0077 0000 006f 0000 0072 0000 006c 0000
00000f0 0064 0000 0021 0000 003c 0000 002f 0000
0000100 006d 0000 0065 0000 0073 0000 0073 0000
0000110 0061 0000 0067 0000 0065 0000 003e 0000
0000120 000a 0000                              
0000124
```

* rexml (3.3.8)
```
$ ruby test_utf.rb 
/Users/naitoh/.rbenv/versions/3.3.4/lib/ruby/gems/3.3.0/gems/rexml-3.3.8/lib/rexml/parsers/baseparser.rb:516:in `pull_event': Malformed XML: Content at the start of the document (got '') (REXML::ParseException)
Line: 2
Position: 292
Last 80 unconsumed characters:
<?xml version="1.0" encoding="UTF-32"?> <
[...]

PR #213, opened 2024-10-09
Support UTF-32 Encoding

naitoh:
GitHub: fix https://github.com/ruby/rexml/issues/212


lib/rexml/parsers/baseparser.rb
```diff
@@ -747,8 +748,12 @@ def process_instruction
           if need_source_encoding_update?(encoding)
             @source.encoding = encoding
           end
-          if encoding.nil? and /\AUTF-16(?:BE|LE)\z/i =~ @source.encoding
-            encoding = "UTF-16"
+          if encoding.nil?
+            if /\AUTF-32(?:BE|LE)\z/i =~ @source.encoding
+              encoding = "UTF-32"
+            elsif encoding.nil? and /\AUTF-16(?:BE|LE)\z/i =~ @source.encoding
```
naitoh, line 754:
```suggestion
            elsif /\AUTF-16(?:BE|LE)\z/i =~ @source.encoding
```
I'll fix it later.

naitoh:
https://github.com/ruby/rexml/issues/212#issuecomment-2412679533

I too will close UTF-32 support as it seems unnecessary.

PR #215, opened 2024-10-18
Fix `IOSource#readline` for `@pending_buffer`

naitoh:
## Why?
Fixed a problem that `@pending_buffer` is not processed when `IOError` occurs in `@source.readline` although `@pending_buffer` exists when reading XML file.


kou:
Thanks.

PR #216, opened 2024-10-21
Add `IOSource#match?` method

naitoh:
## Why?
`StringScanner#match?` is faster than `StringScanner#check`.

See: https://github.com/ruby/strscan/pull/111

## Benchmark
```
RUBYLIB= BUNDLER_ORIG_RUBYLIB= /Users/naitoh/.rbenv/versions/3.3.4/bin/ruby -v -S benchmark-driver /Users/naitoh/ghq/github.com/naitoh/rexml/benchmark/parse.yaml
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [arm64-darwin22]
                 dom     18.819      19.362        32.846       34.708 i/s -     100.000 times in 5.313905s 5.164791s 3.044500s 2.881200s
                 sax     28.188      29.982        48.386       52.554 i/s -     100.000 times in 3.547597s 3.335304s 2.066732s 1.902809s
                pull     31.962      33.902        57.868       60.662 i/s -     100.000 times in 3.128689s 2.949690s 1.728071s 1.648467s
              stream     31.436      33.030        52.808       56.647 i/s -     100.000 times in 3.181095s 3.027574s 1.893635s 1.765304s

Comparison:
                              dom
         after(YJIT):        34.7 i/s
        before(YJIT):        32.8 i/s - 1.06x  slower
               after:        19.4 i/s - 1.79x  slower
              before:        18.8 i/s - 1.84x  slower

                              sax
         after(YJIT):        52.6 i/s
        before(YJIT):        48.4 i/s - 1.09x  slower
               after:        30.0 i/s - 1.75x  slower
              before:        28.2 i/s - 1.86x  slower

                             pull
         after(YJIT):        60.7 i/s
        before(YJIT):        57.9 i/s - 1.05x  slower
               after:        33.9 i/s - 1.79x  slower
              before:        32.0 i/s - 1.90x  slower

                           stream
         after(YJIT):        56.6 i/s
        before(YJIT):        52.8 i/s - 1.07x  slower
               after:        33.0 i/s - 1.72x  slower
              before:        31.4 i/s - 1.80x  slower

```

- YJIT=ON : 1.05x - 1.09x faster
- YJIT=OFF : 1.02x - 1.06x faster

### DIFF
--- a/lib/rexml/source.rb
+++ b/lib/rexml/source.rb
@@ -18,6 +18,16 @@ module REXML
           pattern = /#{Regexp.escape(pattern)}/ if pattern.is_a?(String)
           super(pattern)
         end
+
+        def match?(pattern)
+          pattern = /#{Regexp.escape(pattern)}/ if pattern.is_a?(String)
+          super(pattern)
+        end
+
+        def skip(pattern)
+          pattern = /#{Regexp.escape(pattern)}/ if pattern.is_a?(String)
+          super(pattern)
+        end
       end
     end
     using StringScannerCheckScanString
--- a/lib/rexml/source.rb
+++ b/lib/rexml/source.rb
@@ -136,6 +136,14 @@ module REXML
       end
     end
 
+    def match?(pattern, cons=false)
+      if cons
+        !@scanner.skip(pattern).nil?
+      else
+        !@scanner.match?(pattern).nil?
+      end
+    end
+
     def position
       @scanner.pos
     end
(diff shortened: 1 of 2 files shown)


lib/rexml/source.rb
```diff
@@ -267,6 +285,23 @@ def match( pattern, cons=false )
...
+    def skip( pattern, cons=false )
+      # To avoid performance issue, we need to increase bytes to read per scan
+      min_bytes = 1
+      while true
+        if cons
+          md = @scanner.skip(pattern)
+        else
+          md = @scanner.match?(pattern)
+        end
+        return md if md
```
kou, line 297: `md` -> `n_matched_bytes`

kou (commented):
In general, this is a good improvement.
But can we use better name than `skip`? `skip("...")` doesn't skip.

naitoh:
> But can we use better name than `skip`? `skip("...")` doesn't skip.

I see.
I changed the method name to match_size().

kou:
It seems that all callers don't use the returned size. They just check whether the return value is `nil` or not.

Is `size` in `match_size` a important information? Can we use just a boolean instead of the matched size as a return value instead?

naitoh:
> Is `size` in `match_size` an important information? 

No.

> Can we use just a boolean instead of the matched size as a return value instead?

OK.
I changed the method name to `match?()`.

kou:
Thanks.
`match?` is better than `skip` and `match_size`.

PR #219, opened 2024-10-30
Stop requiring stringio dynamically

jsvd:
`SourceFactory::create_from(String)` will always run the `require 'stringio'` operation. This prevents a multi-threaded JRuby application from parsing xml on separate threads concurrently given that `require` will pass through a synchronized piece of code.

An experiment in removing this `require` lead to a 10x performance improvement on 10 threads parsing incoming strings on xml. For more details see https://github.com/logstash-plugins/logstash-filter-xml/issues/83

### DIFF
--- a/lib/rexml/source.rb
+++ b/lib/rexml/source.rb
@@ -1,8 +1,8 @@
 # coding: US-ASCII
 # frozen_string_literal: false
 
-require "strscan"
 require "stringio"
+require "strscan"
 
 require_relative 'encoding'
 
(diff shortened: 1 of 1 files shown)


lib/rexml/source.rb
```diff
@@ -2,6 +2,7 @@
 # frozen_string_literal: false
 
 require "strscan"
+require "stringio"
```
kou, line 5: Could you sort this list in alphabetical order?

```suggestion
require "stringio"
require "strscan"
```

kou:
Thanks.

PR #221, opened 2024-11-16
Fix error handling when parsing XML via IO.pipe

naitoh:
## Why?

If via IO.pipe, `IOError` exception is not raised, but `Errno::ESPIPE` or `Errno::EPIPE` or `Errno::EINVAL` exception is raised.

- CRuby
```
@er_source.pos
#=> Errno::ESPIPE Exception: Illegal seek
```

- CRuby (Windows (Ruby 2.6 or earlier))
```
@er_source.pos
#=> Errno::EINVAL: Invalid argument
```

- JRuby
```
@er_source.pos
#=> Errno::EPIPE: Broken pipe - No message available
```

### DIFF
--- a/lib/rexml/source.rb
+++ b/lib/rexml/source.rb
@@ -321,7 +321,7 @@ module REXML
         rescue
         end
         @er_source.seek(pos)
-      rescue IOError
+      rescue IOError, SystemCallError
         pos = -1
         line = -1
       end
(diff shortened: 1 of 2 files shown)


lib/rexml/source.rb
```diff
@@ -321,7 +321,7 @@ def current_line
         rescue
         end
         @er_source.seek(pos)
-      rescue IOError
+      rescue IOError, Errno::EPIPE, Errno::ESPIPE, Errno::EINVAL
```
kou, line 324: We can use `SystemCallError` that is an ancestor class of them:

```suggestion
      rescue IOError, SystemCallError
```

naitoh: OK.
I see.

kou:
Thanks.

Issue #228, opened 2025-01-07
REXML::Formatters::Transitive not found.


Beagle123:
Hi:

I tried to use REXML::Formatters::Transitive in my code, but it wasn't found in the package.

Here is the output of irb:

```
eric@trig:~$ irb
irb(main):001:0> require "rexml"
=> true
irb(main):002:0> REXML::Formatters::Transitive
(irb):2:in `<main>': uninitialized constant REXML::Formatters::Transitive (NameError)
        from /home/eric/.rbenv/versions/3.1.6/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'                                                                                                       
        from /home/eric/.rbenv/versions/3.1.6/bin/irb:25:in `load'                                        
        from /home/eric/.rbenv/versions/3.1.6/bin/irb:25:in `<main>'                                      
irb(main):003:0> 
^C                                                                                                        
irb(main):003:0> 
```

I tried using the code from github, Transitive.rb, but the output was bad.  All the tags were broken onto two lines.

However, I had great success with subclassing Pretty.rb, and outputting the text nodes without formatting which is the result that I wanted.  I would suggest making this Formatter:

```
class PrettyPreserveText < REXML::Formatters::Pretty
  def write_text( node, output )
    output << node.to_s()
  end
end
```

This is a great result.  It has all the benefits of Pretty.rb while preserving long text nodes.  It seems like a good idea to preserve text nodes "as-is."  In fact, I would suggest changing the Pretty formater  to output the unaltered text.  When I used Pretty, it worked great except that it would indent every line of a long text string to make it look good.  This obviously is rarely a good idea.  The users need to be able to control the content of text strings.  That's not the job of a formatter.

Thanks for all your work.

Eric

Beagle123:
This is what the output of @doc.write(transitive: true) looks like now:

```

<interface
><requires lib='gtk+' version='3.24'
  /><object class='GtkWindow' id='window1'
  ><property name='default-height'
    >1000</property
    ><property name='default-width'
    >1400</property
    ><property name='modal'
    >False</property
    ><child
    ><object class='GtkBox' id='main_box'
      ><property name='margin-top'
        >20</property
        ><property name='margin-bottom'
        >20</property
        ><property name='margin-start'
        >20</property
        ><property name='margin-end'
        >19</property
        ><property name='spacing'
        >10</property
        ><property name='orientation'
        >vertical</property

```

[...]

PR #229, opened 2025-01-07
Update node.rb  required transitive.rb, so it can be found.

Beagle123:
I recently tried to use REXML::Formatters::Transitive, and its not included in REXML.

### DIFF
--- a/lib/rexml/document.rb
+++ b/lib/rexml/document.rb
@@ -388,7 +388,6 @@ module REXML
       end
       formatter = if indent > -1
           if transitive
-            require_relative "formatters/transitive"
             REXML::Formatters::Transitive.new( indent, ie_hack )
           else
             REXML::Formatters::Pretty.new( indent, ie_hack )
--- a/lib/rexml/document.rb
+++ b/lib/rexml/document.rb
@@ -388,6 +388,7 @@ module REXML
       end
       formatter = if indent > -1
           if transitive
+            
             REXML::Formatters::Transitive.new( indent, ie_hack )
           else
             REXML::Formatters::Pretty.new( indent, ie_hack )
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -1505,7 +1505,7 @@ module REXML
       Kernel.warn("#{self.class.name}.write is deprecated.  See REXML::Formatters", uplevel: 1)
       formatter = if indent > -1
           if transitive
-            require_relative "formatters/transitive"
+            require_relative "formatters/transitive"  
             REXML::Formatters::Transitive.new( indent, ie_hack )
           else
             REXML::Formatters::Pretty.new( indent, ie_hack )
--- a/lib/rexml/node.rb
+++ b/lib/rexml/node.rb
@@ -2,6 +2,7 @@
 require_relative "parseexception"
 require_relative "formatters/pretty"
 require_relative "formatters/default"
+require_relative "formatters/transitive"
 
 module REXML
   # Represents a node in the tree.  Nodes are never encountered except as
--- a/lib/rexml/node.rb
+++ b/lib/rexml/node.rb
@@ -2,7 +2,6 @@
 require_relative "parseexception"
 require_relative "formatters/pretty"
 require_relative "formatters/default"
-require_relative "formatters/transitive"
 
 module REXML
   # Represents a node in the tree.  Nodes are never encountered except as


naitoh:
Thank you for this PR.

@kou

https://docs.ruby-lang.org/ja/latest/class/REXML=3a=3aFormatters=3a=3aTransitive.html (Sorry, this is a Japanese document only.)

> require 'rexml/document'
> require 'rexml/formatters/transitive'

This documentation explicitly requires 'rexml/formatters/transitive'.

Do you think we should be able to use `REXML::Formatters::Transitive.new` with `require "rexml"` as well as `REXML::Formatters::Pretty.new`?

```
irb(main):001> require 'rexml'
=> true
irb(main):002> REXML::Formatters::Pretty.new
=> #<REXML::Formatters::Pretty:0x000000012333db18 @compact=false, @ie_hack=false, @indentation=2, @level=0, @width=80>
irb(main):003> REXML::Formatters::Transitive.new
(irb):3:in '<main>': uninitialized constant REXML::Formatters::Transitive (NameError)
	from <internal:kernel>:168:in 'Kernel#loop'
	from /Users/naitoh/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/irb-1.14.3/exe/irb:9:in '<top (required)>'
	from /Users/naitoh/.rbenv/versions/3.4.1/bin/irb:25:in 'Kernel#load'
	from /Users/naitoh/.rbenv/versions/3.4.1/bin/irb:25:in '<main>'
```

See: #228

kou:
It's required lazy: https://github.com/ruby/rexml/blob/b70388c2638d90ebd2ae471bd85239d8469b8e62/lib/rexml/document.rb#L391

But the approach may be slow on JRuby. See also: https://github.com/ruby/rexml/pull/219

So, I'm OK with requiring it non-lazy. But it should be done in `document.rb` not `node.rb` because it's not used in `node.rb`.

naitoh:
> It's required lazy:
> 
> https://github.com/ruby/rexml/blob/b70388c2638d90ebd2ae471bd85239d8469b8e62/lib/rexml/document.rb#L391
> 
> But the approach may be slow on JRuby. See also: #219
> 
> So, I'm OK with requiring it non-lazy. But it should be done in `document.rb` not `node.rb` because it's not used in `node.rb`.

I see.
Thanks.

@Beagle123
Could you change this PR to conform to the above policy?

Specifically, they are as follows.

1. Remove the following lazy require.
- https://github.com/ruby/rexml/blob/b70388c2638d90ebd2ae471bd85239d8469b8e62/lib/rexml/document.rb#L391
- https://github.com/ruby/rexml/blob/b70388c2638d90ebd2ae471bd85239d8469b8e62/lib/rexml/element.rb#L1508
2. Add non-lazy require in `document.rb`, not `node.rb`.

kou:
Ah, I didn't notice that we have a lazy require in `element.rb` too.
Then `element.rb` is better than `document.rb` because `document.rb` depends on `element.rb`.

naitoh:
> Ah, I didn't notice that we have a lazy require in `element.rb` too. Then `element.rb` is better than `document.rb` because `document.rb` depends on `element.rb`.

OK, I see.

@Beagle123
Could you change this PR to conform to the above policy?

Specifically, they are as follows.

1. Remove the following lazy require.
- https://github.com/ruby/rexml/blob/b70388c2638d90ebd2ae471bd85239d8469b8e62/lib/rexml/document.rb#L391
- https://github.com/ruby/rexml/blob/b70388c2638d90ebd2ae471bd85239d8469b8e62/lib/rexml/element.rb#L1508
2. Add non-lazy require in `element.rb`, not `node.rb`.

Beagle123:
Hi:

I think I've made the changes requested.

However, I don't think my original issue is resolved.  My issue was that using Pretty.rb resulted in my text lines having a bunch of whitespace inserted so that the XML file looks indented, but all my text has drastically changed.  I was not able to require REXML::Formatters::Transitive as it was not found.  I downloaded Transitive.rb, but the output was totally disjointed.  I had to hack Pretty.rb to make it not alter my text, and that works.  But this is not a good situation for users.  I assure you that nobody will be happy with the output from Transitive, and the only choice is to have their text nodes altered by Pretty.rb.  

I've uploaded my new version of Transitive.rb.  I concede that it is a change from the current version, but it does the desireable behavior that Transitive is supposed to do:  Pretty output with preserved text lines.

Thanks,
Eric

lib/rexml/element.rb
```diff
@@ -1505,7 +1505,7 @@ def write(output=$stdout, indent=-1, transitive=false, ie_hack=false)
       Kernel.warn("#{self.class.name}.write is deprecated.  See REXML::Formatters", uplevel: 1)
       formatter = if indent > -1
           if transitive
-            require_relative "formatters/transitive"
+            require_relative "formatters/transitive"
```
naitoh, line 1508: This position is lazy `require`.
Please move this at the beginning of this file so that it is non-lazy `require`.

https://github.com/ruby/rexml/blob/b70388c2638d90ebd2ae471bd85239d8469b8e62/lib/rexml/element.rb#L1-L8

lib/rexml/document.rb
```diff
@@ -388,7 +388,7 @@ def write(*arguments)
       end
       formatter = if indent > -1
           if transitive
-            require_relative "formatters/transitive"
+
```
naitoh, line 391: Please remove this empty line.

naitoh (commented):
> I think I've made the changes requested.

Thanks.
Please make some more corrections.

naitoh:
> However, I don't think my original issue is resolved.

I would like to discuss that issue in #230.

naitoh:
@Beagle123

> > I think I've made the changes requested.
> 
> Thanks. Please make some more corrections.

Is it possible to fix it?

naitoh:
Since I haven't received a reply, I'm closing this PR.

PR #230, opened 2025-01-07
Update transitive.rb

### LINKED ISSUE #228: REXML::Formatters::Transitive not found.

Beagle123:
GitHub: fix GH-228

This makes Transitive.rb do exactly what the docs say it should do.  It outputs Pretty xml except for text nodes which are left untouched.  If you read the issue that I posted, you can see that the current Transitive output is garbled.  Also, this has the benefit of being initialized using Prestty.rb, initialize method for simplicity.

### DIFF
--- a/lib/rexml/formatters/transitive.rb
+++ b/lib/rexml/formatters/transitive.rb
@@ -12,7 +12,7 @@ module REXML
     # Note that this is only useful if the original XML is not already
     # formatted.  Since this formatter does not alter whitespace nodes, the
     # results of formatting already formatted XML will be odd.
-    class Transitive < REXML::Formatters::Pretty
+    class Transitive < Pretty
       def write_text( node, output )
         output << node.to_s()
       end
--- a/lib/rexml/formatters/transitive.rb
+++ b/lib/rexml/formatters/transitive.rb
@@ -12,44 +12,7 @@ module REXML
     # Note that this is only useful if the original XML is not already
     # formatted.  Since this formatter does not alter whitespace nodes, the
     # results of formatting already formatted XML will be odd.
-    class Transitive < Default
-      def initialize( indentation=2, ie_hack=false )
-        @indentation = indentation
-        @level = 0
-        @ie_hack = ie_hack
-      end
-
-      protected
-      def write_element( node, output )
-        output << "<#{node.expanded_name}"
-
-        node.attributes.each_attribute do |attr|
-          output << " "
-          attr.write( output )
-        end unless node.attributes.empty?
-
-        output << "\n"
-        output << ' '*@level
-        if node.children.empty?
-          output << " " if @ie_hack
-          output << "/"
-        else
-          output << ">"
-          # If compact and all children are text, and if the formatted output
-          # is less than the specified width, then try to print everything on
-          # one line
-          @level += @indentation
-          node.children.each { |child|
-            write( child, output )
-          }
-          @level -= @indentation
-          output << "</#{node.expanded_name}"
-          output << "\n"
-          output << ' '*@level
-        end
-        output << ">"
-      end
-
+    class Transitive < REXML::Formatters::Pretty
       def write_text( node, output )
         output << node.to_s()
       end
--- a/lib/rexml/formatters/transitive.rb
+++ b/lib/rexml/formatters/transitive.rb
@@ -14,7 +14,7 @@ module REXML
     # results of formatting already formatted XML will be odd.
     class Transitive < Pretty
       def write_text( node, output )
-        output << node.to_s()
+        output << node.to_s().strip
       end
     end
   end


tompng:
Formatting this html content
```ruby
html = "<html><title> foo </title><body><div> bar </div></body></html>"
REXML::Document.new(html).write(STDOUT, 2, true)
```

In the document
> This means that no extra whitespace nodes are inserted, and whitespace within text nodes is preserved.

This is transitive output. It looks weird, but it's indented, it's valid, and the content is preserved. 
Just like the document, `title.text == ' foo '` and `div.text == ' bar '` is unchanged and no whitespace text node is inserted between `</title>` and `<body>`.
This kind of formatting is the only way to indent without inserting newline nor whitespace between `>` and `<`.
```html
<html
><title
  > foo </title
  ><body
  ><div
    > bar </div
    ></body
  ></html
>
```

This is not transitive. It looks pretty but the content is not preserved.
`title.text == "\nfoo\n  "` and `div.text == "\nbar\n    "` is changed from the original.
Text node `"\n  "` is inserted between `</title>` and `<body>`.
```html
<html>
  <title>
foo
  </title>
  <body>
    <div>
bar
    </div>
  </body>
</html>
```

Beagle123:
If you guys want to preserve Transitive the way that it is for backward compatibility, that's fine with me.  My issue is that I need a user friendly (pretty) output that doesn't change my text nodes.  

I don't see any scenario where having the formatter change the text is advantageous.  When I write my xml with Pretty, it adds a bunch of whitespace so when I read it back it's wrong.  Its a different string that doesn't work.  This makes no sense to me becuase the purpose of XML is to be read by a program, not a human.  Pretty makes it human readable but now program readable.  

I really like Pretty.rb.  I'd love to see it have an option to preserve text nodes.  That would be the best imho.

Are you guys receptive to that instead?

kou:
> In fact, I would suggest making the default behavior of Pretty be to preserve text nodes, and have an option to indent with a bunch of whitespace. That makes sense to me.

What does "text nodes" mean?

Could you provide a sample XML and an expected pretty output for it?

naitoh:
@tompng

> In the document
> 
> > This means that no extra whitespace nodes are inserted, and whitespace within text nodes is preserved.
> 
> This is transitive output. It looks weird, but it's indented, it's valid, and the content is preserved. Just like the document, `title.text == ' foo '` and `div.text == ' bar '` is unchanged and no whitespace text node is inserted between `</title>` and `<body>`. This kind of formatting is the only way to indent without inserting newline nor whitespace between `>` and `<`.

Thanks for your comment.
I understand the purpose of `REXML::Formatters::Transitive`. 

`REXML::Formatters::Transitive` is implemented as documented, and I don't think `REXML::Formatters::Transitive` needs to be changed.

@Beagle123 

https://github.com/ruby/rexml/issues/228
> In fact, I would suggest changing the Pretty formater to output the unaltered text.

Depending on your answer below, I think it would be more appropriate to extend `REXML::Formatters::Pretty`.

>> In fact, I would suggest making the default behavior of Pretty be to preserve text nodes, and have an option to indent with a bunch of whitespace. That makes sense to me.

> What does "text nodes" mean?
> Could you provide a sample XML and an expected pretty output for it?

naitoh:
> > In fact, I would suggest making the default behavior of Pretty be to preserve text nodes, and have an option to indent with a bunch of whitespace. That makes sense to me.
> 
> What does "text nodes" mean?
> 
> Could you provide a sample XML and an expected pretty output for it?

@Beagle123

Could you answer this?

naitoh:
Since I haven't received a reply, I'm closing this PR.

PR #235, opened 2025-01-20
Added rdoc as a development dependency (for Ruby 3.5+)

OlofKalufs:
Ruby 3.5+ requires that rdoc explicitly be declared as a dependency
Should sort out GitHub Actions that are failing due to this


hsbt:
👍

PR #236, opened 2025-01-22
Changed benchmark target to Ruby latest

naitoh:
Ruby 3.4 has been released, we will change our benchmark target to Ruby latest(3.4).

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby-version:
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby-version:
-          - '3.3'
+          - '3.4'
```
kou, line 14: We can use `'ruby'` if we always use the latest Ruby.

kou:
Thanks.

PR #240, opened 2025-02-07
Use test task of Rake instead for testing

gemmaro:
* `Rakefile`: Use `TestTask` of Rake; remains "test" task and now accepts `TEST` environment variable and so on.
* `test/run.rb`: Removed file.
* `.github/workflows/test.yml` (`gem`): Updated to utilize Rake's test task instead of the deleted `run.rb` script.


kou:
I don't like `Rake::TestTask` because it's inconvenient. For example, we can't pass arguments via command line arguments (`test/run.rb -t /TargetTest/`). We need to use environment variable (`rake test TESTOPTS="-t /TargetTest/"`, `rake test TEST=test/XXX.rb`) instead.

gemmaro:
Thank you for sharing your thoughts! I agree that using the existing script is better. I'll close this issue thread.

PR #249, opened 2025-04-26
Improve using `//` in XPath performance

naitoh:
When using `//` in XPath, the deeper the tag hierarchy, the slower it becomes due to the namespace acquisition process.
Caching namespace information improves performance when using `//` with XPath.

## Benchmark (Comparison with rexml 3.4.1)

```
                                                     rexml 3.4.1      master  3.4.1(YJIT)  master(YJIT)
REXML::XPath.match(REXML::Document.new(xml), 'a//a')      29.215     234.909      108.945       492.410 i/s -     100.000 times in 3.422925s 0.425697s 0.917898s 0.203083s

Comparison:
                                        master(YJIT):       492.4 i/s
                                              master:       234.9 i/s - 2.10x  slower
                                         3.4.1(YJIT):       108.9 i/s - 4.52x  slower
                                         rexml 3.4.1:        29.2 i/s - 16.85x  slower
```

- YJIT=ON : 4.52x faster
- YJIT=OFF : 8.04x faster

### DIFF
--- a/lib/rexml/attribute.rb
+++ b/lib/rexml/attribute.rb
@@ -206,6 +206,10 @@ module REXML
       path += "/@#{self.expanded_name}"
       return path
     end
+
+    def document
+      @element&.document
+    end
   end
 end
 #vim:ts=2 sw=2 noexpandtab:
--- a/lib/rexml/document.rb
+++ b/lib/rexml/document.rb
@@ -448,6 +448,20 @@ module REXML
     end
 
     private
+
+    attr_accessor :namespaces_cache
+
+    # New document level cache is created and available in this block.
+    # This API is thread unsafe. Users can't change this document in this block.
+    def enable_cache
+      @namespaces_cache = {}
+      begin
+        yield
+      ensure
+        @namespaces_cache = nil
+      end
+    end
+
     def build( source )
       Parsers::TreeParser.new( source, self ).parse
     end
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -589,10 +589,12 @@ module REXML
     #   d.elements['//c'].namespaces # => {"x"=>"1", "y"=>"2", "z"=>"3"}
     #
     def namespaces
-      namespaces = {}
-      namespaces = parent.namespaces if parent
-      namespaces = namespaces.merge( attributes.namespaces )
-      return namespaces
+      namespaces_cache = document&.__send__(:namespaces_cache)
+      if namespaces_cache
+        namespaces_cache[self] ||= calculate_namespaces
+      else
+        calculate_namespaces
+      end
     end
 
     # :call-seq:
@@ -619,17 +621,9 @@ module REXML
       if prefix.nil?
         prefix = prefix()
       end
-      if prefix == ''
-        prefix = "xmlns"
-      else
-        prefix = "xmlns:#{prefix}" unless prefix[0,5] == 'xmlns'
-      end
-      ns = nil
-      target = self
-      while ns.nil? and target
-        ns = target.attributes[prefix]
-        target = target.parent
-      end
+      prefix = (prefix == '') ? 'xmlns' : prefix.delete_prefix("xmlns:")
+      ns = namespaces[prefix]
+
       ns = '' if ns.nil? and prefix == 'xmlns'
       return ns
     end
@@ -1516,8 +1510,15 @@ module REXML
       formatter.write( self, output )
     end
 
-
     private
+    def calculate_namespaces
+      if parent
+        parent.namespaces.merge(attributes.namespaces)
+      else
+        attributes.namespaces
+      end
+    end
+
     def __to_xpath_helper node
       rv = node.expanded_name.clone
       if node.parent
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -78,7 +78,15 @@ module REXML
 
     def parse path, node
       path_stack = @parser.parse( path )
-      match( path_stack, node )
+      if node.is_a?(Array)
+        Kernel.warn("REXML::XPath.each, REXML::XPath.first, REXML::XPath.match dropped support for nodeset...", uplevel: 1)
+        return [] if node.empty?
+        node = node.first
+      end
+
+      node.document.__send__(:enable_cache) do
+        match( path_stack, node )
+      end
     end
 
     def get_first path, node
@@ -137,11 +145,6 @@ module REXML
 
 
     def match(path_stack, node)
-      if node.is_a?(Array)
-        Kernel.warn("REXML::XPath.each, REXML::XPath.first, REXML::XPath.match dropped support for nodeset...", uplevel: 1)
-        return [] if node.empty?
-        node = node.first
-      end
       nodeset = [XPathNode.new(node, position: 1)]
       result = expr(path_stack, nodeset)
       case result
@@ -494,14 +497,10 @@ module REXML
                 if strict?
                   raw_node.name == name and raw_node.namespace == ""
                 else
-                  # FIXME: This DOUBLES the time XPath searches take
-                  ns = get_namespace(raw_node, prefix)
-                  raw_node.name == name and raw_node.namespace == ns
+                  raw_node.name == name and raw_node.namespace == get_namespace(raw_node, prefix)
                 end
               else
-                # FIXME: This DOUBLES the time XPath searches take
-                ns = get_namespace(raw_node, prefix)
-                raw_node.name == name and raw_node.namespace == ns
+                raw_node.name == name and raw_node.namespace == get_namespace(raw_node, prefix)
               end
             when :attribute
               if prefix.nil?
@@ -509,9 +508,7 @@ module REXML
               elsif prefix.empty?
                 raw_node.name == name and raw_node.namespace == ""
               else
-                # FIXME: This DOUBLES the time XPath searches take
-                ns = get_namespace(raw_node.element, prefix)
-                raw_node.name == name and raw_node.namespace == ns
+                raw_node.name == name and raw_node.namespace == get_namespace(raw_node.element, prefix)
               end
             else
               false
--- a/test/test_core.rb
+++ b/test/test_core.rb
@@ -653,17 +653,22 @@ Last 80 unconsumed characters:
       assert_equal "<sean:blah>Some text</sean:blah>", out
     end
 
-
     def test_add_namespace
       e = Element.new 'a'
+      assert_equal("", e.namespace)
+      assert_nil(e.namespace('foo'))
       e.add_namespace 'someuri'
       e.add_namespace 'foo', 'otheruri'
       e.add_namespace 'xmlns:bar', 'thirduri'
-      assert_equal 'someuri', e.attributes['xmlns']
-      assert_equal 'otheruri', e.attributes['xmlns:foo']
-      assert_equal 'thirduri', e.attributes['xmlns:bar']
+      assert_equal("someuri", e.namespace)
+      assert_equal("otheruri", e.namespace('foo'))
+      assert_equal("otheruri", e.namespace('xmlns:foo'))
+      assert_equal("thirduri", e.namespace('bar'))
+      assert_equal("thirduri", e.namespace('xmlns:bar'))
+      assert_equal('someuri', e.attributes['xmlns'])
+      assert_equal('otheruri', e.attributes['xmlns:foo'])
+      assert_equal('thirduri', e.attributes['xmlns:bar'])
     end
-
 
     def test_big_documentation
       d = File.open(fixture_path("documentation.xml")) {|f| Document.new f }
@@ -764,9 +769,15 @@ Last 80 unconsumed characters:
 
     def test_delete_namespace
       doc = Document.new "<a xmlns='1' xmlns:x='2'/>"
+      assert_equal("1", doc.root.namespace)
+      assert_equal("2", doc.root.namespace('x'))
+      assert_equal("2", doc.root.namespace('xmlns:x'))
       doc.root.delete_namespace
       doc.root.delete_namespace 'x'
-      assert_equal "<a/>", doc.to_s
+      assert_equal("<a/>", doc.to_s)
+      assert_equal("", doc.root.namespace)
+      assert_nil(doc.root.namespace('x'))
+      assert_nil(doc.root.namespace('xmlns:x'))
     end
 
     def test_each_element_with_attribute
--- a/test/xpath/test_base.rb
+++ b/test/xpath/test_base.rb
@@ -1193,6 +1193,16 @@ EOF
       assert_equal( 1,  XPath.match( d, "//x:*" ).size )
     end
 
+    def test_namespaces_cache
+      doc = Document.new("<a xmlns='1'><b/></a>")
+      assert_equal("<b/>", XPath.first(doc, "//b[namespace-uri()='1']").to_s)
+      assert_nil(XPath.first(doc, "//b[namespace-uri()='']"))
+
+      doc.root.delete_namespace
+      assert_nil(XPath.first(doc, "//b[namespace-uri()='1']"))
+      assert_equal("<b/>", XPath.first(doc, "//b[namespace-uri()='']").to_s)
+    end
+
     def test_ticket_71
       doc = Document.new(%Q{<root xmlns:ns1="xyz" xmlns:ns2="123"><element ns1:attrname="foo" ns2:attrname="bar"/></root>})
       el = doc.root.elements[1]
--- a/benchmark/xpath.yaml
+++ b/benchmark/xpath.yaml
@@ -1,0 +1,32 @@
+loop_count: 100
+contexts:
+  - gems:
+      rexml: 3.2.6
+    require: false
+    prelude: require 'rexml'
+  - name: master
+    prelude: |
+      $LOAD_PATH.unshift(File.expand_path("lib"))
+      require 'rexml'
+  - name: 3.2.6(YJIT)
+    gems:
+      rexml: 3.2.6
+    require: false
+    prelude: |
+      require 'rexml'
+      RubyVM::YJIT.enable
+  - name: master(YJIT)
+    prelude: |
+      $LOAD_PATH.unshift(File.expand_path("lib"))
+      require 'rexml'
+      RubyVM::YJIT.enable
+
+prelude: |
+  require 'rexml/document'
+
+  DEPTH = 100
+  xml   = '<a>' * DEPTH + '</a>' * DEPTH
+  doc   = REXML::Document.new(xml)
+
+benchmark:
+  "REXML::XPath.match(REXML::Document.new(xml), 'a//a')" : REXML::XPath.match(doc, "a//a")


lib/rexml/element.rb
```diff
@@ -621,17 +621,9 @@ def namespace(prefix=nil)
...
-      ns = nil
-      target = self
-      while ns.nil? and target
-        ns = target.attributes[prefix]
-        target = target.parent
-      end
-      ns = '' if ns.nil? and prefix == 'xmlns'
-      return ns
+
+      namespaces[prefix].to_s
```
tompng, line 626: It's changed to always returns a string, so the method comment needs to be updated
```diff
- #   b.namespace('nosuch') # => nil
+ #   b.namespace('nosuch') # => ""
```

lib/rexml/element.rb
```diff
@@ -589,10 +589,10 @@ def prefixes
     #   d.elements['//c'].namespaces # => {"x"=>"1", "y"=>"2", "z"=>"3"}
     #
     def namespaces
-      namespaces = {}
-      namespaces = parent.namespaces if parent
-      namespaces = namespaces.merge( attributes.namespaces )
-      return namespaces
+      @namespaces ||= parent.namespaces if parent
+      @namespaces ||= {}
+      @namespaces = @namespaces.merge( attributes.namespaces )
```
tompng, line 594: I think this `merge` part should also be cached and only called once, because cache pattern normally caches the whole calculation of the value to be cached.

```ruby
def namespaces
  # cache whole `@namespaces` calculation
  @namespaces ||= (parent ? parent.namespaces : {}).merge(attributes.namespaces)
  @namespaces.dup
end

# cache parent_namespaces pattern
def namespaces
  @parent_namespaces ||= parent ? parent.namespaces : {}
  @parent_namespaces.merge(attributes.namespaces)
end
```

lib/rexml/element.rb
naitoh: Oh, sorry.
This is an incompatible change, so I will change it to return nil.
Thanks!

lib/rexml/element.rb
naitoh: `Element#add_namespace` and `Element#delete_namespace` methods change the `namespaces` information.

Therefore, it is not a good idea to cache the entire result of the `@namespaces` calculation.
Also, add a process to clear the cache when the `namespaces` information of the parent node is updated, since this will affect the calculation of the child nodes.

 Thanks!

lib/rexml/element.rb
```diff
@@ -1516,6 +1508,11 @@ def write(output=$stdout, indent=-1, transitive=false, ie_hack=false)
       formatter.write( self, output )
     end
 
+    protected
+    def reset_parent_namespaces
+      @parent_namespaces = nil
+      self.each{|e| e.reset_parent_namespaces }
```
tompng, line 1514: Looks like TextNode needs to be handled correctly
```ruby
doc = REXML::Document.new('<a><a>a</a></a>')
doc.add_namespace 'foo'
#=> undefined method 'reset_parent_namespaces' for an instance of REXML::Text
```

tompng:
I have some concerns about maintainability because cache invalidation bug is hard to debug.
Namespace can be modified through add_attribute/delete_attribute and through `attr_reader :attributes`. It can bypass cache invalidation mechanism.

Here are some ideas of cache strategy.

1. Use a temporary cache and pass it as a method parameter

```ruby
def node_test
  namespaces_cache = {}
  # Cache scope is enclosed in this method.
  # There is no need to worry about cache invalidation.
  call node.namespaces(namespaces_cache) to use the cache
end

def namespaces(cache = nil)
  if cache
    cache[self] ||= calculate_namespaces(cache)
  else
    calculate_namespaces
  end
end
```

2. Automatically invalidate cache with immutable structure

```ruby
def namespaces
  parent_namespaces = parent&.namespaces
  # If parent.namespaces or @attributes is changed, cache is automatically invalidated
  return @namespaces if parent_namespaces.equal?(@parent_namespaces) && @attrubutes.frozen?

  @attributes.freeze # Need to unfreeze before modifying @attributes
  @parent_namespaces = parent_namespaces
  @namespaces = @parent_namespaces.merge(@attributes).freeze
end
```

lib/rexml/element.rb
```diff
@@ -653,6 +643,7 @@ def namespace(prefix=nil)
     #   e.namespaces # => {"xmlns"=>"bar", "baz"=>"bat"}
     #
     def add_namespace( prefix, uri=nil )
+      self.each{|e| e.reset_parent_namespaces }
```
kou, line 646:
```suggestion
      @elements.each(&:reset_parent_namespaces)
```

kou: BTW, if there are many descendant elements, is `add_namespace` slow?

lib/rexml/element.rb
naitoh: Oh sorry.
I will fix this case.

naitoh:
> I have some concerns about maintainability because cache invalidation bug is hard to debug. Namespace can be modified through add_attribute/delete_attribute and through `attr_reader :attributes`. It can bypass cache invalidation mechanism.
> 
> Here are some ideas of cache strategy.
> 
>     1. Use a temporary cache and pass it as a method parameter
> 
> 
> ```ruby
> def node_test
>   namespaces_cache = {}
>   # Cache scope is enclosed in this method.
>   # There is no need to worry about cache invalidation.
>   call node.namespaces(namespaces_cache) to use the cache
> end
> 
> def namespaces(cache = nil)
>   if cache
>     cache[self] ||= calculate_namespaces(cache)
>   else
>     calculate_namespaces
>   end
> end
> ```

For conservatism, I have changed my mind about adopting policy 1.
Thanks!

kou:
I have a bit concern about adding a `cache` (positional) argument to only some public namespace related APIs.

I feel that it may a bit ad-hoc approach. I'm not object this approach but can we improve our API?

For example, how about keeping a cache per document?

```ruby
class XPathParser
  def parse path, nodeset
    nodeset.first.document.enable_cache do
      # New document level cache is created and available in this block.
      match( path_stack, nodeset )
    end
  end
end
```

naitoh:
I add `Document#enable_namespaces_cache` method to control whether caching is enabled or disabled.
What do you think?

lib/rexml/element.rb
```diff
@@ -589,10 +589,12 @@ def prefixes
     #   d.elements['//c'].namespaces # => {"x"=>"1", "y"=>"2", "z"=>"3"}
     #
     def namespaces
-      namespaces = {}
-      namespaces = parent.namespaces if parent
-      namespaces = namespaces.merge( attributes.namespaces )
-      return namespaces
+      namespaces_cache = root.document&.namespaces_cache
```
kou, line 592: Do we need `root` here?

```suggestion
      namespaces_cache = document&.namespaces_cache
```

lib/rexml/document.rb
```diff
@@ -443,6 +443,17 @@ def record_entity_expansion
...
+
+    def enable_namespaces_cache
+      self.namespaces_cache = {}
+      begin
+        yield
+      ensure
+        self.namespaces_cache = nil
+      end
+    end
+
```
kou, line 456: Can we hide this API or make this API internal?

lib/rexml/document.rb
```diff
@@ -443,6 +443,17 @@ def record_entity_expansion
       end
     end
 
+    attr_accessor :namespaces_cache
+
+    def enable_namespaces_cache
```
kou, line 448: How about removing `_namespaces`?
I don't want to add more similar `enable_XXX_cache` internal APIs in the future...

lib/rexml/xpath_parser.rb
```diff
@@ -78,7 +78,15 @@ def variables=( vars={} )
 
     def parse path, nodeset
       path_stack = @parser.parse( path )
-      match( path_stack, nodeset )
+      if nodeset.first.node_type == :attribute
```
kou, line 81: Could you care `nodeset.empty?` case?

Can we remove this check by adding `Attribute#document`?

lib/rexml/document.rb
naitoh: OK
I see.

lib/rexml/xpath_parser.rb
```diff
@@ -78,7 +78,15 @@ def variables=( vars={} )
...
+        match( path_stack, nodeset )
+      else
+        nodeset.first.document.enable_namespaces_cache do
+          # New document level cache is created and available in this block.
+          # This API is thread unsafe. Users can't change this document in this block.
```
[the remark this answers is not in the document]
naitoh: OK
I see.

lib/rexml/xpath_parser.rb
naitoh: OK
I see.

naitoh:
I have corrected the points you pointed out.
What do you think?

lib/rexml/document.rb
```diff
@@ -447,7 +447,21 @@ def document
       self
     end
 
+    attr_accessor :namespaces_cache
```
kou, line 450: Can we hide this?

(Hmm. Is adding a cache related API to document a bad idea...?)

lib/rexml/xpath_parser.rb
```diff
@@ -78,7 +78,9 @@ def variables=( vars={} )
 
     def parse path, nodeset
       path_stack = @parser.parse( path )
-      match( path_stack, nodeset )
+      nodeset.first&.document&.send(:enable_cache) do
```
kou, line 81: If the first node doesn't have a document, this returns `nil`. Is this backward compatible?

lib/rexml/document.rb
naitoh: OK
I see.

naitoh:
I have corrected the points you pointed out.
What do you think?

lib/rexml/xpath_parser.rb
tompng: `XPath.match`, `XPath.first` and `XPath.each` can accept both node and array, so this is possible.
```ruby
REXML::XPath.match([] ,'*')
REXML::XPath.first([] ,'*')
REXML::XPath.each([] ,'*'){}
```

This usage is not documented, but used in test codes.
https://github.com/ruby/rexml/blob/5d2606a99957f1623e464d70be065a31ed0ea58b/test/xpath/test_base.rb#L415
https://github.com/ruby/rexml/blob/5d2606a99957f1623e464d70be065a31ed0ea58b/test/test_jaxen.rb#L86

Currently, size of the array passed to match/first/each in the test code is always 1.

naitoh: Oh sorry...
I added `nodeset.empty?` test case.

> Similarly, I would expect an error if the first node does not have a document object.

I added a test case where the first node does not have a document object.

naitoh:
I have corrected the points you pointed out.
What do you think?

naitoh:
I have corrected the points you pointed out.
What do you think?

kou:
Let's revisit this after #252. This implementation will be changed by #252.

naitoh:
#252 merged, so I rebased this PR and updated.
What do you think?

lib/rexml/attribute.rb
```diff
@@ -206,6 +206,10 @@ def xpath
       path += "/@#{self.expanded_name}"
       return path
     end
+
+    def document
+      self.element.document
```
kou, line 211:
```suggestion
      @element&.document
```

lib/rexml/document.rb
```diff
@@ -448,6 +448,20 @@ def document
     end
 
     private
+
+    attr_accessor :namespaces_cache
+
+    # New document level cache is created and available in this block.
+    # This API is thread unsafe. Users can't change this document in this block.
+    def enable_cache
+      self.namespaces_cache = {}
```
kou, line 457:
```suggestion
      @namespaces_cache = {}
```

lib/rexml/document.rb
```diff
@@ -448,6 +448,20 @@ def document
...
+      begin
+        yield
+      ensure
+        self.namespaces_cache = nil
```
kou, line 461:
```suggestion
        @namespaces_cache = nil
```

lib/rexml/element.rb
```diff
@@ -589,10 +589,12 @@ def prefixes
     #   d.elements['//c'].namespaces # => {"x"=>"1", "y"=>"2", "z"=>"3"}
     #
     def namespaces
-      namespaces = {}
-      namespaces = parent.namespaces if parent
-      namespaces = namespaces.merge( attributes.namespaces )
-      return namespaces
+      namespaces_cache = document&.send(:namespaces_cache)
```
kou, line 592:
```suggestion
      namespaces_cache = document&.__send__(:namespaces_cache)
```

lib/rexml/element.rb
```diff
@@ -1516,8 +1510,11 @@ def write(output=$stdout, indent=-1, transitive=false, ie_hack=false)
       formatter.write( self, output )
     end
 
-
     private
+    def calculate_namespaces
+      (parent ? parent.namespaces : {}).merge(attributes.namespaces)
```
kou, line 1515:
```suggestion
      (parent&.namespaces || {}).merge(attributes.namespaces)
```

kou: Or:

```suggestion
      if parent
        parent.namespaces.merge(attributes.namespaces)
      else
        attributes.namespaces
      end
```

lib/rexml/xpath_parser.rb
```diff
@@ -78,7 +78,19 @@ def variables=( vars={} )
 
     def parse path, node
       path_stack = @parser.parse( path )
-      match( path_stack, node )
+      if node.is_a?(Array)
```
kou, line 81: Can we remove this check because we dropped support for `Array` here?

naitoh: @kou 

If this check is remove, the following deprecated message will no longer be output...

https://github.com/ruby/rexml/blob/cd575a10cac58eb47f235ed186060ac65ffb5284/lib/rexml/xpath_parser.rb#L139-L144

kou: Can we avoid duplicated `Array` checks?

tompng: I added the deprecation check to `xpath_parser.rb`, but maybe I should have added the deprecation directly to `xpath.rb`'s `def XPath::first`, `def XPath::each` and `def XPath::match`.

(To do that, we need to be careful about the return value especially for `XPath.each(e,p) #=> Enumerator` and `XPath.each(e,p){} #=> Array`)

lib/rexml/document.rb
naitoh: OK.
I see.

lib/rexml/element.rb
naitoh: OK.
I see.

lib/rexml/element.rb
naitoh: OK.
I see.

lib/rexml/attribute.rb
naitoh: OK.
I see.

naitoh:
I have corrected the points you pointed out.
What do you think?

test/xpath/test_base.rb
```diff
@@ -100,6 +100,12 @@ def test_root
       assert_equal "a", XPath::first( doc, "/" ).elements[1].name
     end
 
+    def test_nil_doc
+      assert_raise(NoMethodError) {
+        XPath.match(nil, '*')
+      }
+    end
```
tompng, line 107: I don't think this test is needed.
I think it's fine if the wrong usage `XPath.match(anything_that_is_not_an_element, '*')` raises any kind of error or returns anything without error.

tompng: Other part of this pull request looks good for to :+1:

naitoh: OK
I see.
Thansk!

naitoh:
I have corrected the points you pointed out.
What do you think?

test/test_core.rb
```diff
@@ -653,18 +653,23 @@ def test_namespace
       assert_equal "<sean:blah>Some text</sean:blah>", out
     end
 
-
     def test_add_namespace
       e = Element.new 'a'
+      assert_equal "", e.namespace
+      assert_equal nil, e.namespace('foo')
```
kou, line 659: Could you avoid omitting parentheses for assertions to avoid potential warning?
See also: https://bugs.ruby-lang.org/issues/20922

```suggestion
      assert_equal("", e.namespace)
      assert_equal(nil, e.namespace('foo'))
```

lib/rexml/xpath_parser.rb
```diff
@@ -78,7 +78,15 @@ def variables=( vars={} )
...
+        Kernel.warn("REXML::XPath.each, REXML::XPath.first, REXML::XPath.match dropped support for nodeset...", uplevel: 1)
+        return [] if node.empty?
+        node = node.first
+      end
+
+      node.document.__send__(:enable_cache) do
+        match( path_stack, node )
+      end
```
kou, line 89: Can we add a test for this cache?

For example:
1. It uses `XPathParser#parse`
2. It changes namespace
3. It uses `XPathParser#parse` (this should not use cache used for 1.)

test/test_core.rb
naitoh: OK.
I see.

lib/rexml/xpath_parser.rb
naitoh: OK.
I see.

naitoh:
I have corrected the points you pointed out.
What do you think?

test/xpath/test_base.rb
```diff
@@ -1193,6 +1193,16 @@ def test_namespaces_0
       assert_equal( 1,  XPath.match( d, "//x:*" ).size )
     end
 
+    def test_namespaces_cache
+      doc = Document.new "<a xmlns='1'><b/></a>"
```
kou, line 1197:
```suggestion
      doc = Document.new("<a xmlns='1'><b/></a>")
```

test/xpath/test_base.rb
```diff
@@ -1193,6 +1193,16 @@ def test_namespaces_0
...
+      assert_equal("<b/>", XPath.first( doc, "//b[namespace-uri()='1']" ).to_s)
+      assert_nil(XPath.first( doc, "//b[namespace-uri()='']" ))
```
kou, line 1199:
```suggestion
      assert_nil(XPath.first(doc, "//b[namespace-uri()='']"))
```

test/xpath/test_base.rb
```diff
@@ -1193,6 +1193,16 @@ def test_namespaces_0
...
+
+      doc.root.delete_namespace
+      assert_nil(XPath.first( doc, "//b[namespace-uri()='1']" ))
+      assert_equal("<b/>", XPath.first( doc, "//b[namespace-uri()='']" ).to_s)
```
kou, line 1203: (Do we need the `"//b[namespace-uri()='1']"` case?)

```suggestion
      assert_nil(XPath.first(doc, "//b[namespace-uri()='1']"))
      assert_equal("<b/>", XPath.first(doc, "//b[namespace-uri()='']").to_s)
```

test/xpath/test_base.rb
naitoh: OK.
I see.

test/xpath/test_base.rb
naitoh: OK.
I see.

test/xpath/test_base.rb
naitoh: > (Do we need the `"//b[namespace-uri()='1']"` case?)

I think this is necessary to see changes in cache status.

naitoh:
I have corrected the points you pointed out.
What do you think?

kou:
Thanks.

PR #251, opened 2025-04-30
Fix reverse sort in xpath_parser

tompng:
The code below was failing with `REXML::XPathParser#sort': undefined method '-@' for an instance of Array`
```ruby
d = REXML::Document.new("<a><b><c/><d/><x/></b><b><e/><x/></b></a>")
matches = REXML::XPath.match(d, "a/b/x/preceding-sibling::node()")
# Before: error
# After: [<e/>, <d/>, <c/>]
```
This pull request will fix it.

### DIFF
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -671,7 +671,7 @@ module REXML
         if order == :forward
           index
         else
-          -index
+          index.map(&:-@)
         end
       end
       ordered.collect do |_index, node|
--- a/test/xpath/test_base.rb
+++ b/test/xpath/test_base.rb
@@ -409,6 +409,10 @@ module REXMLTests
       assert_equal("c", matches[0].name)
       assert_equal("b", matches[1].name)
 
+      d = REXML::Document.new("<a><b/><c/><d/><d/></a>")
+      cs = REXML::XPath.match(d, "a/d/preceding::node()")
+      assert_equal(["d", "c", "c", "b", "b"], cs.map(&:name))
+
       s = "<a><b><c id='1'/></b><b><b><c id='2'/><c id='3'/></b><c id='4'/></b><c id='NOMATCH'><c id='5'/></c></a>"
       d = REXML::Document.new(s)
       c = REXML::XPath.match( d, "//c[@id = '5']")
--- a/test/xpath/test_base.rb
+++ b/test/xpath/test_base.rb
@@ -409,15 +409,17 @@ module REXMLTests
       assert_equal("c", matches[0].name)
       assert_equal("b", matches[1].name)
 
-      d = REXML::Document.new("<a><b/><c/><d/><d/></a>")
-      cs = REXML::XPath.match(d, "a/d/preceding::node()")
-      assert_equal(["d", "c", "c", "b", "b"], cs.map(&:name))
-
       s = "<a><b><c id='1'/></b><b><b><c id='2'/><c id='3'/></b><c id='4'/></b><c id='NOMATCH'><c id='5'/></c></a>"
       d = REXML::Document.new(s)
       c = REXML::XPath.match( d, "//c[@id = '5']")
       cs = REXML::XPath.match( c, "preceding::c" )
       assert_equal( 4, cs.length )
+    end
+
+    def test_preceding_sibling
+      d = REXML::Document.new("<a><b><c/><d/><x/></b><b><e/><x/></b></a>")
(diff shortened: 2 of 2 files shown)


naitoh:
Thansk.
However, there may be other problems, since more <d/> tags would increase the number of <c/> and <b/> tags.

## After

```ruby
> $LOAD_PATH.unshift(File.expand_path("lib"))
> require 'rexml'
> REXML::XPath.match(REXML::Document.new("<a><b/><c/><d/></a>"), "/a/d/preceding::node()")
=> [<c/>, <b/>]
> REXML::XPath.match(REXML::Document.new("<a><b/><c/><d/><d/></a>"), "/a/d/preceding::node()")
=> [<d/>, <c/>, <c/>, <b/>, <b/>]
> REXML::XPath.match(REXML::Document.new("<a><b/><c/><d/><d/><d/></a>"), "/a/d/preceding::node()")
=> [<d/>, <d/>, <d/>, <c/>, <c/>, <c/>, <b/>, <b/>, <b/>]
> REXML::XPath.match(REXML::Document.new("<a><b/><c/><d/><d/><d/><d/></a>"), "/a/d/preceding::node()")
=> [<d/>, <d/>, <d/>, <d/>, <d/>, <d/>, <c/>, <c/>, <c/>, <c/>, <b/>, <b/>, <b/>, <b/>]
```

tompng:
I think the duplicated match result is not related to this change.
To avoid adding a test case of questionable duplicated match result, I changed the test to use `preceding-sibling`.
```ruby
d = REXML::Document.new("<a><b><c/><d/><x/></b><b><e/><x/></b></a>")
matches = REXML::XPath.match(d, "a/b/x/preceding-sibling::node()")
# Before: REXML::XPathParser#sort': undefined method '-@' for an instance of Array
# After: [<e/>, <d/>, <c/>]
```

naitoh:
Thanks.
But, the response order is reversed from `nokogiri` and `libxml`, `python(lxml)`.

@kou 
Do you think the rexml responses is in the correct order? 🤔

```ruby
$ irb -rnokogiri
> puts Nokogiri::HTML("<a><b><c/><d/><x/></b><b><e/><x/></b></a>").xpath( "//a/b/x/preceding-sibling::node()")
<c></c>
<d></d>
<e></e>
```

```ruby
$ irb -rlibxml
> puts LibXML::XML::Document.string("<a><b><c/><d/><x/></b><b><e/><x/></b></a>").find("//a/b/x/preceding-sibling::node()").to_a
<c/>
<d/>
<e/>
```

https://programiz.pro/ide/python
```python
from lxml import etree
print(etree.fromstring("<a><b><c/><d/><x/></b><b><e/><x/></b></a>").xpath('//a/b/x/preceding-sibling::node()'))

[<Element c at 0x7887f2cc7580>, <Element d at 0x7887f2cc75c0>, <Element e at 0x7887f2cc7600>]
```

https://scrapinghub.github.io/xpath-playground/




https://www.w3.org/TR/2017/REC-xpath-31-20170321/#doc-xpath31-ReverseAxis

> the preceding-sibling axis contains the context node's preceding siblings, those children of the context node's parent that occur before the context node in document order; 

I don't think it specifically states that the order should be reversed.

kou:
> But, the response order is reversed from `nokogiri` and `libxml`, `python(lxml)`.

All of them are bindings of libxml2. So it's natural that all of them have the same behavior. :-)

> Do you think the rexml responses is in the correct order? 🤔

Yes.

`match` returns a "node-set" that is unordered.

https://www.w3.org/TR/1999/REC-xpath-19991116/#section-Introduction

> node-set (an unordered collection of nodes without duplicates)

FYI: Most implementations use XPath 1.0 not XPath 2.0, 3.1, ... REXML also uses XPath 1.0.

naitoh:
> > But, the response order is reversed from `nokogiri` and `libxml`, `python(lxml)`.
> 
> All of them are bindings of libxml2. So it's natural that all of them have the same behavior. :-)

Oh!
I see.

> > Do you think the rexml responses is in the correct order? 🤔
> 
> Yes.
> 
> `match` returns a "node-set" that is unordered.
> 
> https://www.w3.org/TR/1999/REC-xpath-19991116/#section-Introduction
> 
> > node-set (an unordered collection of nodes without duplicates)
> 
> FYI: Most implementations use XPath 1.0 not XPath 2.0, 3.1, ... REXML also uses XPath 1.0.

Thanks.
I see.

@tompng 
Can you update this PR's description?

tompng:
@naitoh Updated :+1:

naitoh:
Thanks!

PR #252, opened 2025-04-30
Deprecate accepting array as an element in XPath.match, first and each

tompng:
`XPath.match`, `XPath.first`, `XPath.each`, `XPathParser#parse` and `XPathParser#match` accepted nodeset as element.
This pull request changes the first parameter of these method to be an element instead of nodeset.
Passing nodeset will be deprecated.

```ruby
# Documented usage. OK
REXML::XPath.match(element, xpath)

# Undocumented usage. Deprecate in this pull request
nodeset = [element]
REXML::XPath.match(nodeset, xpath)
```

### Background
#249 will introduce a temporary cache.
```ruby
def parse path, nodeset
  nodeset.first.document.send(:enable_cache) do
    match( path_stack, nodeset )
  end
end
```
But the signature `XPathParser#match(path, nodeset)` does not guarantee that all nodes in the nodeset has the same root document.
So cache does not work in the code below. It's still slow.
```ruby
REXML::XPath.match(2.times.map { REXML::Document.new('<a>'*400+'</a>'*400) }, 'a//a')
```

The interface is holding our back, so I propose to drop accepting array as element.
This change is a backward incompatibility, but it just drops undocumented feature. I think only the test code was unintentionally using this feature.

### XPath.match with array
XPath.match only traverse the first element of the array for some selectors.
```ruby
nodeset = [REXML::Document.new("<a><b/></a>"), REXML::Document.new("<a><c/></a>")]

REXML::XPath.match(nodeset, "a/*")
#=> [<b/>, <c/>]

REXML::XPath.match(nodeset, "//a/*")
#=> [<b/>] # I expect [<b/>, <c/>] but the second document is ignored
```
It indicates that `XPath.match` is not designed to search inside multiple nodes/documents.

### DIFF
--- a/test/test_jaxen.rb
+++ b/test/test_jaxen.rb
@@ -56,7 +56,9 @@ module REXMLTests
 
     # processes a tests/document/context node
     def process_context(doc, context)
-      test_context = XPath.match(doc, context.attributes["select"])
+      matched = XPath.match(doc, context.attributes["select"])
+      assert_equal(matched.size, 1)
+      test_context = matched.first
       namespaces = context.namespaces
       namespaces.delete("var")
       namespaces = nil if namespaces.empty?
@@ -104,7 +106,8 @@ module REXMLTests
       end
 
       XPath.each(test, "valueOf") do |value_of|
-        process_value_of(matched, variables, namespaces, value_of)
+        assert_equal(1, matched.size)
+        process_value_of(matched.first, variables, namespaces, value_of)
       end
     end
 
@@ -118,10 +121,8 @@ module REXMLTests
 
     def user_message(context, xpath, matched)
       message = ""
-      context.each_with_index do |node, i|
-        message << "Node#{i}:\n"
-        message << "#{node}\n"
-      end
+      message << "Node:\n"
+      message << "#{context}\n"
       message << "XPath: <#{xpath}>\n"
       message << "Matched <#{matched}>"
       message
--- a/test/test_jaxen.rb
+++ b/test/test_jaxen.rb
@@ -103,6 +103,8 @@ module REXMLTests
         assert_equal(Integer(expected, 10),
                      matched.size,
                      user_message(context, xpath, matched))
+      else
+        assert_operator(matched.size, :>, 0, user_message(context, xpath, matched))
       end
 
       XPath.each(test, "valueOf") do |value_of|
--- a/test/test_jaxen.rb
+++ b/test/test_jaxen.rb
@@ -57,7 +57,7 @@ module REXMLTests
     # processes a tests/document/context node
     def process_context(doc, context)
       matched = XPath.match(doc, context.attributes["select"])
-      assert_equal(matched.size, 1)
+      assert_equal(1, matched.size)
       test_context = matched.first
       namespaces = context.namespaces
       namespaces.delete("var")
@@ -106,8 +106,9 @@ module REXMLTests
       end
 
       XPath.each(test, "valueOf") do |value_of|
-        assert_equal(1, matched.size)
-        process_value_of(matched.first, variables, namespaces, value_of)
+        matched.each do |subcontext|
+          process_value_of(subcontext, variables, namespaces, value_of)
+        end
       end
     end
 
--- a/test/xpath/test_base.rb
+++ b/test/xpath/test_base.rb
@@ -411,9 +411,10 @@ module REXMLTests
 
       s = "<a><b><c id='1'/></b><b><b><c id='2'/><c id='3'/></b><c id='4'/></b><c id='NOMATCH'><c id='5'/></c></a>"
       d = REXML::Document.new(s)
-      c = REXML::XPath.match( d, "//c[@id = '5']")
-      cs = REXML::XPath.match( c, "preceding::c" )
-      assert_equal( 4, cs.length )
+      c = REXML::XPath.match(d, "//c[@id = '5']")
+      assert_equal(1, c.length)
+      cs = REXML::XPath.match(c.first, "preceding::c")
+      assert_equal(4, cs.length)
     end
 
     def test_following
--- a/test/xpath/test_base.rb
+++ b/test/xpath/test_base.rb
@@ -1159,5 +1159,15 @@ EOF
       end
       assert_equal(["/"], hrefs, "Bug #3842 [ruby-core:32447]")
     end
+
+    def test_match_with_deprecated_usage
+      verbose, $VERBOSE = $VERBOSE, nil
+      doc = Document.new("<a><b/></a>")
+      assert_equal(['b'], XPath.match([doc, doc], '//b').map(&:name))
+      assert_equal(['b'], XPath.match([doc], '//b').map(&:name))
+      assert_equal([], XPath.match([], '//b').map(&:name))
+    ensure
+      $VERBOSE = verbose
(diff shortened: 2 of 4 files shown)


naitoh:
> But the signature `XPathParser#match(path, nodeset)` does not guarantee that all nodes in the nodeset has the same root document. So cache does not work in the code below. It's still slow.

In this case, I made further improvements to #249, which eliminated the slowness.

> Drop accepting array as an element in XPath.match, first and each

How about instead of removing a feature, a deprecated message should be displayed if it is specified in an array?

What do you think @kou?

test/test_jaxen.rb
```diff
@@ -104,7 +106,8 @@ def process_nominal_test(context, variables, namespaces, test)
       end
 
       XPath.each(test, "valueOf") do |value_of|
-        process_value_of(matched, variables, namespaces, value_of)
+        assert_equal(1, matched.size)
+        process_value_of(matched.first, variables, namespaces, value_of)
```
kou, line 110:
```suggestion
        matched.each do |subcontext|
          process_value_of(subcontext, variables, namespaces, value_of)
        end
```

kou (commented):
I'm OK that we drop support for nodeset because we don't have enough resource to complete nodeset support.

In general, we want to keep backward compatibility as much as possible. But we can remove the feature without keeping backward compatibility because:

1. It's not documented
2. It doesn't work in some cases

But could you report a warning as @naitoh suggested something like:

```diff
diff --git a/lib/rexml/xpath_parser.rb b/lib/rexml/xpath_parser.rb
index 5eb1e5a..a2b2ef5 100644
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -136,11 +136,12 @@ module REXML
     end
 
 
-    def match(path_stack, nodeset)
-      nodeset = nodeset.collect.with_index do |node, i|
-        position = i + 1
-        XPathNode.new(node, position: position)
+    def match(path_stack, node)
+      if node.is_a?(Array)
+        warn("REXML::XPath.XXX dropped support for nodeset...", uplevel: N)
+        node = node.first
       end
+      nodeset = [XPathNode.new(node, position: 1)]
       result = expr(path_stack, nodeset)
       case result
       when Array # nodeset
```

test/test_jaxen.rb
tompng: Instead of assuming test target of `<valueOf select="attr">value</valueOf>` to be exactly `1`, I add an assertion of `count > 0`.

```xml
<test select="foo/bar" count="N"> <!-- expected value of matched.size is N -->
  <valueOf select="attr">value</valueOf>
</test>
<test select="foo/bar"> <!-- matched.size should be larger than zero -->
  <valueOf select="attr">value</valueOf>
</test>
```
If the `count` is not specified, and `matched.size` is zero, we're missing a potential bug.

test/xpath/test_base.rb
```diff
@@ -1158,5 +1159,15 @@ def test_or_and
       end
       assert_equal(["/"], hrefs, "Bug #3842 [ruby-core:32447]")
     end
+
+    def test_match_with_deprecated_usage
+      verbose, $VERBOSE = $VERBOSE, nil
+      doc = Document.new("<a><b/></a>")
+      assert_equal ['b'], XPath.match([doc, doc], '//b').map(&:name)
+      assert_equal ['b'], XPath.match([doc], '//b').map(&:name)
+      assert_equal [], XPath.match([], '//b').map(&:name)
```
kou, line 1168: Could you avoid omitting parenthesis for assertions to avoid potential warnings?
See also: https://bugs.ruby-lang.org/issues/20922

```suggestion
      assert_equal(['b'], XPath.match([doc, doc], '//b').map(&:name))
      assert_equal(['b'], XPath.match([doc], '//b').map(&:name))
      assert_equal([], XPath.match([], '//b').map(&:name))
```

kou (approved):

naitoh (approved):
LGTM
Thanks!

naitoh:
@tompng
Can you update this PR's description?

tompng:
@naitoh 
Updated. Change the title `Drop` → `Deprecate` and mention about deprecation in the description

naitoh:
Thanks!

naitoh:
@voxik 

> Just FTR, this likely breaks vagrant-libvirt:
 
> I have no idea ATM how to fix this :/

We apologize for causing incompatibility with vagrant-libvirt.
To avoid the impact of this change, please implement the following countermeasures at the location where the message `REXML::XPath.each, REXML::XPath.first, REXML::XPath.match dropped support for nodeset...` appears.

* Please specify a single node, not an array of nodes, for REXML::XPath.match.
* When specifying an array of nodes to receive multiple results, process the nodes using `each` and specify a single node.

> https://github.com/vagrant-libvirt/vagrant-libvirt/blob/a94ce0d7b6c90129a38435698ca97b364130313d/lib/vagrant-libvirt/action/resolve_disk_settings.rb#L53-L62

### Fixed example:

```ruby
$ git diff -ub
diff --git a/lib/vagrant-libvirt/action/resolve_disk_settings.rb b/lib/vagrant-libvirt/action/resolve_disk_settings.rb
index f1a3f67..bf5d951 100644
--- a/lib/vagrant-libvirt/action/resolve_disk_settings.rb
+++ b/lib/vagrant-libvirt/action/resolve_disk_settings.rb
@@ -51,14 +51,16 @@ module VagrantPlugins
               xml_descr = REXML::Document.new(domain_xml)
               domain_name = xml_descr.elements['domain'].elements['name'].text
               disks_xml = REXML::XPath.match(xml_descr, '/domain/devices/disk[@device="disk"]')
-              have_aliases = !REXML::XPath.match(disks_xml, './alias[@name="ua-box-volume-0"]').first.nil?
+              have_aliases = !REXML::XPath.match(disks_xml.first, './alias[@name="ua-box-volume-0"]').first.nil?
               env[:ui].warn(I18n.t('vagrant_libvirt.domain_xml.obsolete_method')) unless have_aliases
 
               if have_aliases
-                REXML::XPath.match(disks_xml,
+                disks_xml.each do |disk_xml|
+                  REXML::XPath.match(disk_xml,
                                      './alias[contains(@name, "ua-box-volume-")]').each_with_index do |alias_xml, idx|
                     domain_volumes.push(volume_from_xml(alias_xml.parent, domain_name, idx))
                   end
+                end
               else
                 # fallback to try and infer which boxes are box images, as they are listed first
                 # as soon as there is no match, can exit

```

voxik:
> ### Fixed example:

Thanks, very useful to see real life example from experts 👍

Since I have proposed slightly different fix in vagrant-libvirt/vagrant-libvirt#1861 would you mind to check it and elaborate there what are the pros and cons of your vs my proposal?

voxik:
> Each time, since we specify the XPath from the top of the XML, the number of processing steps increases. Therefore, when the target XML is large, there may be a performance impact.

That was also my worry, right. Thanks for confirming.

On top of that, I think your proposal uncovers flaw in the logic which was earlier elaborated in the original description: _"XPath.match only traverse the first element of the array for some selectors."_, because I does not seem to me it was intention to look only at first "disk". But I am not an author of the code, so I'll leave the decision to the upstream.

Thx for the suggestion and elaborating 👍 I appreciate that.

PR #253, opened 2025-05-02
NEWS.md : Fix the mentioned of the PR in CVE-2024-35176.

naitoh:
I think the mentioned of CVE-2024-35176 in NEWS.md is incorrect.

```
- Improved parse performance when an attribute has many <s.
  - GH-126
```

#126 looks like fixes the issue with attribute value that contains multiple '>' characters.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -386,7 +386,7 @@
 
     * Patch by NAITOH Jun.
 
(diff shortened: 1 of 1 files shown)


kou:
Right.

PR #255, opened 2025-05-04
fix: XPath following, following-sibling, preceding, preceding-sibling  for multiple node

naitoh:
## Why?
See: https://github.com/ruby/rexml/pull/251#issuecomment-2845103143

- XPath : a/d/preceding::* => ["d", "c", "b"]
```xml
<a>
  <b/> <!-- a/d/preceding::b -->
  <c/> <!-- a/d/preceding::c -->
  <d/> <!-- a/d/preceding::d -->
  <d/> <!-- self -->
  <e/>
  <f/>
</a>
```

- XPath : a/d/following::* => ["d", "e", "f"]
```xml
<a>
  <b/>
  <c/>
  <d/> <!-- self -->
  <d/> <!-- a/d/following::d -->
  <e/> <!-- a/d/following::e -->
  <f/> <!-- a/d/following::f -->
</a>
```

- XPath : a/b/x/following-sibling:* => ["c", "d", "e"]
```xml
<a>
  <b>
    <x/> <!-- self -->
    <c/> <!-- a/b/x/following-sibling::c -->
    <d/> <!-- a/b/x/following-sibling::d -->
  </b>
  <b>
    <x/> <!-- self -->
    <e/> <!-- a/b/x/following-sibling::e -->
  </b>
</a>
```

- XPath : a/b/x/following-sibling:* => ["c", "d", "x", "e"]
```xml
<a>
  <b>
    <x/> <!-- self -->
    <c/> <!-- a/b/x/following-sibling::c -->
    <d/> <!-- a/b/x/following-sibling::d -->
    <x/> <!-- a/b/x/following-sibling::x -->
    <e/> <!-- a/b/x/following-sibling::e -->
  </b>
</a>
```

- XPath : a/b/x/preceding-sibling::* => ["e", "d", "c"]
```xml
<a>
  <b>
    <c/>  <!-- a/b/x/preceding-sibling::c -->
    <d/>  <!-- a/b/x/preceding-sibling::d -->
    <x/>  <!-- self -->
  </b>
  <b>
    <e/>  <!-- a/b/x/preceding-sibling::e -->
    <x/>  <!-- self -->
  </b>
</a>
```

- XPath : a/b/x/preceding-sibling::* => ["e", "x", "d", "c"]
```xml
<a>
  <b>
    <c/>  <!-- a/b/x/preceding-sibling::c -->
    <d/>  <!-- a/b/x/preceding-sibling::d -->
    <x/>  <!-- a/b/x/preceding-sibling::x -->
    <e/>  <!-- a/b/x/preceding-sibling::e -->
    <x/>  <!-- self -->
  </b>
</a>
```


tompng:
It has a breaking change.
```ruby
d = REXML::Document.new('<div>
  <div>
    <a/><w/>
  </div>
  <a/><x/>
  <a/><y/>
  <a/><z/>
</div>')
# Finds a node flowing <a/>
REXML::XPath.match(d, '//a/following-sibling::*[1]')
# Before: [<w/>, <x/>, <y/>, <z/>]
# After: [<w/>, <x/>]
```

For reference, the behavior in master branch is same as JavaScript xpath.

```javascript
document.body.innerHTML = `
<div>
  <div><a></a><w></w></div>
  <a></a><x></x>
  <a></a><y></y>
  <a></a><z></z>
</div>`
result = document.evaluate('//a/following-sibling::*[1]', document, null, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE, null)
for(i=0;i<result.snapshotLength;i++) console.log(i, result.snapshotItem(i))
// 0 <w></w>
// 1 <x></x>
// 2 <y></y>
// 3 <z></z>
```

naitoh:
> It has a breaking change.

Thanks!
I have corrected the points pointed out by removing duplications.

lib/rexml/xpath_parser.rb
```diff
@@ -147,7 +147,7 @@ def match(path_stack, nodeset)
         unnode(result)
       else
         [result]
-      end
+      end.uniq
```
kou, line 150: Do we need `.uniq` for the `[result]` case?

test/xpath/test_base.rb
```diff
@@ -416,10 +416,49 @@ def test_preceding
...
+    end
+
     def test_preceding_sibling
       d = REXML::Document.new("<a><b><c/><d/><x/></b><b><e/><x/></b></a>")
       matches = REXML::XPath.match(d, "a/b/x/preceding-sibling::node()")
       assert_equal(["e", "d", "c"], matches.map(&:name))
+
+      d = REXML::Document.new("<a><b><c/><d/><x/><e/><x/></b></a>")
+      matches = REXML::XPath.match(d, "a/b/x/preceding-sibling::node()")
+      assert_equal(["e", "x", "d", "c"], matches.map(&:name))
```
kou, line 461: Could you use a separated test with suitable test name?
What is the important difference with the existing `test_preceding_sibling` assertion?

(BTW, how about indenting the target XML for easy to read?)

test/xpath/test_base.rb
```diff
@@ -416,10 +416,49 @@ def test_preceding
...
+        <div>
+          <a/><w/>
+        </div>
+        <a/><x/>
+        <a/><y/>
+        <a/><z/>
+      </div>")
+      # Finds a node flowing <a/>
+      matches = REXML::XPath.match(d, "//a/following-sibling::*[1]")
+      assert_equal(["w", "x", "y", "z"], matches.map(&:name))
```
kou, line 447: Could you use a separated test with suitable test name?

test/xpath/test_base.rb
```diff
@@ -416,10 +416,49 @@ def test_preceding
...
+
+      d = REXML::Document.new("<a><b><x/><c/><d/><x/><e/></b></a>")
+      matches = REXML::XPath.match(d, "a/b/x/following-sibling::node()")
+      assert_equal(["c", "d", "x", "e"], matches.map(&:name))
```
kou, line 451: Could you use a separated test with suitable test name?

lib/rexml/xpath_parser.rb
naitoh: In the case of `[result]`, it is not necessary.
I would remove it.
Thanks!

test/xpath/test_base.rb
naitoh: OK.
I see.

test/xpath/test_base.rb
naitoh: OK.
I see.

test/xpath/test_base.rb
naitoh: > Could you use a separated test with suitable test name? 

OK.
I see.

> What is the important difference with the existing `test_preceding_sibling` assertion?

- XPath : a/b/x/preceding-sibling::* => ["e", "d", "c"] # existing `test_preceding_sibling`
```xml
<a>
  <b>
    <c/>  <!-- a/b/x/preceding-sibling::c -->
    <d/>  <!-- a/b/x/preceding-sibling::d -->
    <x/>  <!-- self -->
  </b>
  <b>
    <e/>  <!-- a/b/x/preceding-sibling::e -->
    <x/>  <!-- self -->
  </b>
</a>
```

- XPath : a/b/x/preceding-sibling::* => ["e", "x", "d", "c"] # new
```xml
<a>
  <b>
    <c/>  <!-- a/b/x/preceding-sibling::c -->
    <d/>  <!-- a/b/x/preceding-sibling::d -->
    <x/>  <!-- a/b/x/preceding-sibling::x -->
    <e/>  <!-- a/b/x/preceding-sibling::e -->
    <x/>  <!-- self -->
  </b>
</a>
```

The test cases are added because the results differ depending on whether multiple nodes are included in a single node or multiple nodes.

> (BTW, how about indenting the target XML for easy to read?)

OK.
I see.

naitoh:
I have corrected the points you pointed out.
What do you think?

test/xpath/test_base.rb
```diff
@@ -416,12 +416,89 @@ def test_preceding
       assert_equal( 4, cs.length )
     end
 
-    def test_preceding_sibling
-      d = REXML::Document.new("<a><b><c/><d/><x/></b><b><e/><x/></b></a>")
-      matches = REXML::XPath.match(d, "a/b/x/preceding-sibling::node()")
+    def test_preceding_multiple
+      d = REXML::Document.new("
```
kou, line 420: How about using here-document like existing tests?

naitoh: OK.
I see.

test/xpath/test_base.rb
```diff
@@ -416,12 +416,103 @@ def test_preceding
...
-    def test_preceding_sibling
-      d = REXML::Document.new("<a><b><c/><d/><x/></b><b><e/><x/></b></a>")
-      matches = REXML::XPath.match(d, "a/b/x/preceding-sibling::node()")
+    def test_preceding_multiple
+      source = <<-XML
+<a>
+  <b/><c/><d/><d/><e/><f/>
+</a>
+      XML
+      d = REXML::Document.new( source )
```
kou, line 425: * Let's use `XXX(YYY)` style not `XXX( YYY )` style for newly added codes
* Let's use `doc` or `document` instead of `d` because we have `<d/>` in this context

```suggestion
      doc = REXML::Document.new(source)
```

test/xpath/test_base.rb
```diff
@@ -416,12 +416,103 @@ def test_preceding
...
+      matches = REXML::XPath.match(d, "a/d/preceding::*")
```
kou, line 426:
```suggestion
      matches = REXML::XPath.match(doc, "a/d/preceding::*")
```

test/xpath/test_base.rb
naitoh: OK.
I see.

naitoh:
I have corrected the points you pointed out.
What do you think?

kou:
Thanks.

PR #258, opened 2025-07-07
Fix wrong Encoding resolution

kou:
In this context, `Encoding` means `REXML::Encoding` not `Encoding`.


nobu:
There may be another issue.
`REXML::Encoding#find_encoding` may return a lower-case name if `name` has lower-case, and `REXML::Encoding#encoding=` compares the previous `@encoding` with it before `upcase`.
In that case, it returns `true` unexpectedly (probably).
Usually the return value of an assignment method affects nothing, but `REXML::Source#encoding=` calls `super`, so `encoding_updated` will be called even the encoding is unchanged actually.

Issue #260, opened 2025-07-07
NoMethodError: undefined method 'enable_cache' for nil


pboling:
I have started testing libraries against various released versions of their dependencies, as well as the git HEAD version of the dependencies so I will have advance notice when things break.

This gem works fine with ruby-openid / ruby-openid2 in all released versions I've tested, but on git HEAD I am seeing the following error:

```
NoMethodError: undefined method 'enable_cache' for nil
    /var/home/pboling/.local/share/mise/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/bundler/gems/rexml-3dc9eca877f8/lib/rexml/xpath_parser.rb:87:in 'REXML::XPathParser#parse'
    /var/home/pboling/.local/share/mise/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/bundler/gems/rexml-3dc9eca877f8/lib/rexml/xpath.rb:66:in 'REXML::XPath.each'
    /var/home/pboling/.local/share/mise/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/bundler/gems/rexml-3dc9eca877f8/lib/rexml/element.rb:1965:in 'REXML::Elements#each'
```

It is occuring in CI here:
https://github.com/ruby-openid/ruby-openid2/actions/runs/16096185925/job/45419317075#step:6:49

To reproduce, after checking out the repo:
```
git clone https://github.com/ruby-openid/ruby-openid2.git
cd ruby-openid2
git checkout next
BUNDLE_GEMFILE=gemfiles/dep_heads.gemfile bundle install
TEST=test/test_filters.rb BUNDLE_GEMFILE=gemfiles/dep_heads.gemfile bundle exec rake test
```

pboling:
Running against a local checkout of the `rexml` project:
```
.......[81, 90] in /var/home/pboling/src/ruby/rexml/lib/rexml/xpath_parser.rb
    81|       if node.is_a?(Array)
    82|         Kernel.warn("REXML::XPath.each, REXML::XPath.first, REXML::XPath.match dropped support for nodeset...", uplevel: 1)
    83|         return [] if node.empty?
    84|         node = node.first
    85|       end
=>  86|       binding.b
    87|       node.document.__send__(:enable_cache) do
    88|         match( path_stack, node )
    89|       end
    90|     end
[...]

PR #261, opened 2025-07-07
🐛 Properly handle Element#document # => nil

pboling:
- Fixes https://github.com/ruby/rexml/issues/260

### DIFF
--- a/lib/rexml/child.rb
+++ b/lib/rexml/child.rb
@@ -83,13 +83,12 @@ module REXML
     # Returns:: the document this child belongs to, or nil if this child
     # belongs to no document
     def document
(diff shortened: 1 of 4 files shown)


kou (commented):
Could you split this PR to logically separated PRs?

For example:

* Refactoring by `&.` (This must not change the current behavior)
* Caching retrieved `doctype` in a local variable (This must not change the current behavior)
* Removing redundant `return` (This must not change the current behavior)
* Reusing `XPath.match` (This must not change the current behavior)
* Changing `XPath.match` behavior
* Adding support for `Elements#each` without document

pboling:
Sure!

PR #265, opened 2025-07-08
♻️ Use Safe Navigation (&.) from Ruby 2.3

pboling:
Split out from #261, and improved

- [x] No changes in behavior


kou:
Thanks.

PR #266, opened 2025-07-08
♻️ Remove redundant return statements

pboling:
Split from #261; expanded and improved.

- [x] Very slight behavior change here in `REXML::Valdiation::Event#matches?` , which is to align the predicate method's return value with the expected behavior of a predicate method (which is to return one of true or false).  See comments for more details.

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -206,12 +206,12 @@ module REXML
 
       # Returns true if there are no more events
       def empty?
-        return (@source.empty? and @stack.empty?)
+        (@source.empty? and @stack.empty?)
       end
 
       # Returns true if there are more events.  Synonymous with !empty?
       def has_next?
-        return !(@source.empty? and @stack.empty?)
+        !(@source.empty? and @stack.empty?)
       end
 
       # Push an event back on the head of the stream.  This method
@@ -522,7 +522,6 @@ module REXML
           raise REXML::ParseException.new( "Exception parsing",
             @source, self, (error ? error : $!) )
         end
-        return [ :dummy ]
       end
       private :pull_event
 
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -522,6 +522,8 @@ module REXML
           raise REXML::ParseException.new( "Exception parsing",
             @source, self, (error ? error : $!) )
         end
+        # NOTE: The end of the method never runs, because it is unreachable.
+        #       All branches of code above have explicit unconditional return or raise statements.
       end
       private :pull_event
 
--- a/lib/rexml/quickpath.rb
+++ b/lib/rexml/quickpath.rb
@@ -41,7 +41,7 @@ module REXML
       else
         results = filter([element], path)
       end
-      return results
+      results
     end
 
     # Given an array of nodes it filters the array based on the path. The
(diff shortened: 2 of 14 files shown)


lib/rexml/parsers/baseparser.rb
```diff
@@ -522,7 +522,6 @@ def pull_event
           raise REXML::ParseException.new( "Exception parsing",
             @source, self, (error ? error : $!) )
         end
-        return [ :dummy ]
```
pboling, line 525: NOTE: This was unreachable code. No execution path could possibly get here, so the statement was entirely superfluous.

As such it was untested, being literally impossible to test, and there is no behavior change.

lib/rexml/validation/relaxng.rb
```diff
@@ -262,9 +262,10 @@ def next( event )
           rv = super
           return rv if rv
           @prior = @previous.pop
-          return @prior.next( event )
+          @prior.next( event )
+        else
+          super
         end
-        super
```
pboling, line 267: Former construction was odd, since it could be mistakenly read as calling `super` twice. Keeping `super` in the else statement prevents this misreading.

No behavior change.

lib/rexml/quickpath.rb
```diff
@@ -98,9 +98,10 @@ def QuickPath::filter elements, path
           #     results |= filter( children, $' )
           #end
         end
-        return results
+        results
+      else
+        []
       end
-      return []
```
pboling, line 103: This allows a lack of clarity around potential fall through of unhandled cases.  Using the explicit `else` makes the code more declarative.

No behavior change.

lib/rexml/validation/validation.rb
```diff
@@ -80,26 +80,26 @@ def done?
...
 
       def matches?( event )
         return false unless event[0] == @event_type
         case event[0]
         when nil
-          return true
+          true
         when :start_element
-          return true if event[1] == @event_arg
+          event[1] == @event_arg
```
pboling, line 92: This is a slight behavior change, in that this would previously return `nil` when the condition was *not* met, since it would exit the `case` statement without returning early, and exit the method without any explicit return, thus defaulting to void / `nil` return value.

This seems like it was unintentional, and thankfully, since this method appears to be intended as a predicate method, which should always return `true` or `false`, callers are likely using the boolean-ish result, where `nil` is falsey.

Returning actual `false` here then should be backwards compatible with the explicit intent of the `?` predicate method.

lib/rexml/validation/validation.rb
```diff
@@ -80,26 +80,26 @@ def done?
...
         when :end_element
-          return true
+          true
         when :start_attribute
-          return true if event[1] == @event_arg
+          event[1] == @event_arg
```
pboling, line 96: Same as above comment.

Returning actual `false` here, instead of `nil` after falling out of the `case` statement, should be backwards compatible with the explicit intent of the ? predicate method.

lib/rexml/parsers/baseparser.rb
kou: OK. How about using `raise REXML::ParseException.new("Must not happen")` or something?

pboling: I don't see the point, since it is unreachable code.  That line of logic is not accessible in any scenario.

pboling: What I mean is, there would be no way to write a reasonable test for it, since it can't be executed.  All branches of the method have explicit unconditional `return` statements or `raise` errors, and the final line is never reached.

pboling: If the concern is that the method may change in the future and we want to return a loud fail to indicate that we missed handling a certain scenario in the complex logic above, that's a legit concern, but I think it would be better to restructure to logic above instead.

kou: We want to keep the information that this line must not be reached. Other approaches:

1. Add a comment
2. Remove needless `return`s in the large `if`: https://github.com/ruby/rexml/blob/63f3e9772595a64b036953f0ab026d2ea5560a3b/lib/rexml/parsers/baseparser.rb#L432-L516

pboling: Done.

PR #268, opened 2025-07-08
🐛 Handle nil when parsing fragment

### LINKED ISSUE #267: XPathParser#parse should work when node.document is nil
Currently it will raise a `NoMethodError`.
```
    def parse path, node
      if node.is_a?(Array)
        Kernel.warn("REXML::XPath.each, REXML::XPath.first, REXML::XPath.match dropped support for nodeset...", uplevel: 1)
        return [] if node.empty?
        node = node.first
      end

      node.document.__send__(:enable_cache) do # <==== Will fail here
        match( path_stack, node )
      end
    end
```

Instead it should do:

```
if node.document
  node.document.__send__(:enable_cache) do
    match( path_stack, node )
  end
else
  match( path_stack, node )
end
```

This supercedes #260

pboling:
Fixes #267

### DIFF
--- a/test/test_xpath_parser.rb
+++ b/test/test_xpath_parser.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: false
+#
+#  Created by Henrik Mårtensson on 2007-02-18.
+#  Copyright (c) 2007. All rights reserved.
+
+module REXMLTests
+  class TestXPathParser < Test::Unit::TestCase
+    def setup
+      @root_element = _make_service_element(["urn:type1", "urn:type2"], ["http://uri"])
+      @root = @root_element.root
+      @element = @root_element.children[0]
+    end
+
+    def _make_service_element(types, uris)
+      root_element = REXML::Element.new
+      element = root_element.add_element("Service")
+      types.each do |type_text|
+        element.add_element("Type").text = type_text
+      end
+      uris.each do |uri_text|
+        element.add_element("URI").text = uri_text
+      end
+      root_element
+    end
+
+    def test_is_first_child
+      assert_kind_of(REXML::Element, @element)
+    end
+
+    def test_has_element_as_parent
+      assert_kind_of(REXML::Element, @element.parent)
+    end
+
+    def test_has_element_as_root
+      assert_kind_of(REXML::Element, @root)
+    end
+
+    def test_parent_is_root
+      assert_equal(@root, @element.parent)
+    end
+
+    def test_has_nil_siblings
+      assert_nil(@root.previous_sibling)
+      assert_nil(@element.next_sibling)
+    end
+
+    def test_has_not_nil_siblings
+      assert_kind_of(REXML::Element, @element.children[0].next_sibling)
+      assert_kind_of(REXML::Element, @element.children[1].previous_sibling)
+    end
+
+    def test_parsing_when_found
+      @parser = REXML::XPathParser.new
+      res = @parser.parse("/Service", @root_element)
+      assert_equal("[<Service> ... </>]",
+                   res.to_s)
+    end
+
+    def test_parsing_when_not_found
+      @parser = REXML::XPathParser.new
+      res = @parser.parse("/Stardust", @root_element)
+      assert_equal("[]",
+                   res.to_s)
+    end
+  end
+end
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -84,7 +84,12 @@ module REXML
         node = node.first
       end
 
-      node.document.__send__(:enable_cache) do
+      document = node.document
+      if document
+        document.__send__(:enable_cache) do
+          match( path_stack, node )
+        end
+      else
         match( path_stack, node )
       end
     end
--- a/test/parser/test_xpath.rb
+++ b/test/parser/test_xpath.rb
@@ -4,7 +4,7 @@ require "test/unit"
 require "rexml/parsers/xpathparser"
 
 module REXMLTests
-  class TestXPathParser < Test::Unit::TestCase
+  class TestParserXPathParser < Test::Unit::TestCase
     sub_test_case("#abbreviate") do
       def abbreviate(xpath)
         parser = REXML::Parsers::XPathParser.new


test/test_xpath_parser.rb
```diff
@@ -0,0 +1,66 @@
...
+
+module REXMLTests
+  class TestXPathParser < Test::Unit::TestCase
+    def setup
+      @root_element = _make_service_element(["urn:type1", "urn:type2"], ["http://uri"])
+      @root = @root_element.root
+      @element = @root_element.children[0]
+    end
+
+    def _make_service_element(types, uris)
```
kou: We don't need to prepend `_` for private method because we aren't using Python:

```suggestion
    def make_service_element(types, uris)
```

test/test_xpath_parser.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: false
+#
+#  Created by Henrik Mårtensson on 2007-02-18.
+#  Copyright (c) 2007. All rights reserved.
```
kou: You didn't write these tests, right?

test/test_xpath_parser.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def test_has_nil_siblings
+      assert_nil(@root.previous_sibling)
+      assert_nil(@element.next_sibling)
+    end
+
+    def test_has_not_nil_siblings
+      assert_kind_of(REXML::Element, @element.children[0].next_sibling)
+      assert_kind_of(REXML::Element, @element.children[1].previous_sibling)
+    end
```
kou: Why do we need these tests? These tests are for test data, right?

test/test_xpath_parser.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def test_parsing_when_found
+      @parser = REXML::XPathParser.new
+      res = @parser.parse("/Service", @root_element)
+      assert_equal("[<Service> ... </>]",
+                   res.to_s)
```
kou: Could you use an object (`[@root]`) instead of `#to_s` for strict comparision?

test/test_xpath_parser.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    def test_parsing_when_not_found
+      @parser = REXML::XPathParser.new
+      res = @parser.parse("/Stardust", @root_element)
```
kou: How about using more "not found" like element name for easy to read?

```suggestion
      res = @parser.parse("/nonexistent", @root_element)
```

test/test_xpath_parser.rb
```diff
@@ -0,0 +1,66 @@
...
+      assert_nil(@root.previous_sibling)
+      assert_nil(@element.next_sibling)
+    end
+
+    def test_has_not_nil_siblings
+      assert_kind_of(REXML::Element, @element.children[0].next_sibling)
+      assert_kind_of(REXML::Element, @element.children[1].previous_sibling)
+    end
+
+    def test_parsing_when_found
```
kou: `parsing_when` is redundant because this is a `XPathParser` test case.

```suggestion
    def test_found
```

test/test_xpath_parser.rb
```diff
@@ -0,0 +1,66 @@
...
+      @parser = REXML::XPathParser.new
+      res = @parser.parse("/Service", @root_element)
+      assert_equal("[<Service> ... </>]",
+                   res.to_s)
+    end
+
+    def test_parsing_when_not_found
```
kou: Ditto.

test/test_xpath_parser.rb
pboling: Oh, actually I copied existing ones, and modified them.  Should I update the copyright?

test/test_xpath_parser.rb
pboling: Just to ensure the test data is structured as expected, yeah.

test/test_xpath_parser.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    def test_found
+      @parser = REXML::XPathParser.new
+      res = @parser.parse("/Service", @root_element)
+      assert_equal("[<Service> ... </>]",
+                   res.to_s)
+    end
+
+    def test_parsing_when_not_found
```
pboling:
```suggestion
    def test_not_found
```

test/test_xpath_parser.rb
kou: If you don't object, we want to use "Copyright (C) 1993-2013 Yukihiro Matsumoto" mentioned in https://github.com/ruby/rexml/blob/master/LICENSE.txt .

If we don't mention copyright explicitly here, "Copyright (C) 1993-2013 Yukihiro Matsumoto" is used.

test/test_xpath_parser.rb
kou: OK. We want to focus on `XPathParser` tests here. Could you remove them? If we have tests for tests, it's difficult to focus `XPathParser` tests.

test/test_xpath_parser.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: false
+#
+#  Created by Henrik Mårtensson on 2007-02-18.
+#  Copyright (c) 2007. All rights reserved.
+
+module REXMLTests
+  class TestXPathParser < Test::Unit::TestCase
+    def setup
+      @root_element = _make_service_element(["urn:type1", "urn:type2"], ["http://uri"])
+      @root = @root_element.root
```
kou: What is the difference of `@root_element` and `@root`?
In general, root is an XML document but `@root` isn't an XML document, right?

test/test_xpath_parser.rb
pboling: Removed

test/test_xpath_parser.rb
pboling: Done.

test/test_xpath_parser.rb
kou: Thanks.

Note: This assumes that copied code (copyrighted code in an existing test) will be removed finally.

test/test_xpath_parser.rb
pboling: Fixed.

test/test_xpath_parser.rb
pboling: In this case there is no document.  This is testing the behavior with a fragment, i.e. just an Element, that has some child elements.  In this case @root looks like a strange UNDEFINED element.  I wasn't sure if I should add a test for that.  It doesn't seem to be relevant to the XPathParser in particular.

test/test_xpath_parser.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def test_found
+      res = @parser.parse("/Service", @root_element)
+      assert_equal([@element],
+                   res)
+    end
+
+    def test_not_found
+      @parser = REXML::XPathParser.new
```
kou:
```suggestion
```

test/test_xpath_parser.rb
pboling: But also, I've removed `@root`, as it is just confusing, and after removing the redundant tests, it was unsused.

Issue #271, opened 2025-07-08
Add devcontainer for working with Ruby 2.5


pboling:
Many modern computers are not able to install Ruby 2.5 directly, and must resort to using containers.  The [devcontainers standard](https://containers.dev/) can help solve this gap.

I propose adding a devcontainer to the project that is configured to run Ruby 2.5 so we can easily test things on the oldest supported Ruby.

tompng:
CI for Ruby 2.5 is failing for about a month, and there's 990 insertions and 503 deletions that aren't confirmed to pass test in Ruby 2.5.
Perhaps it's time to drop Ruby 2.5 support instead of fixing the CI and/or adding dev container.

kou:
OK. Let's drop support for Ruby 2.5.

PR #272, opened 2025-07-08
🧑‍💻 devcontainer for debugging with Ruby 2.5

pboling:
- [x] Implements #271 

<sub>

Please help me support open source by [sponsoring me](https://github.com/sponsors/pboling) - @pboling

</sub>


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/ruby
+{
+  "name": "Ruby 2.5 on Debian",
+  "image": "ruby:2.5-slim-buster", // Specify the Ruby 2.5 image
```
kou: How about using non slim version for  development?

```suggestion
  "image": "ruby:2.5-buster", // Specify the Ruby 2.5 image
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+  "postCreateCommand": "RUBYGEMS_VERSION=3.3.27\ngem install rubygems-update -v ${RUBYGEMS_VERSION} > /dev/null 2>&1\nupdate_rubygems > /dev/null 2>&1\n", // Optional: Run this command after the container is created
```
kou: Do we need this?
Can we use already installed RubyGems?

BTW, can we run `bundle install` instead to install dependencies automatically?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+
+  // Features to add to the dev container. More info: https://containers.dev/features.
+  "features": {
+    "./apt-install": {}
+  },
```
kou: Do we need this?
It seems that `install.sh` installs needless packages.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "ruby --version",
```
kou: Could you remove needless comments?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+
+	// Configure tool-specific properties.
+  "customizations" : {
+    "jetbrains" : {
+      "backend" : "RubyMine"
+    }
+  },
+
+	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+	// "remoteUser": "root"
```
kou: Could you remove needless comments?

```suggestion
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+    "./apt-install": {}
+  },
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "ruby --version",
+
+	// Configure tool-specific properties.
```
kou: Could you fix indentation?

```suggestion
  // Configure tool-specific properties.
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+  "customizations" : {
+    "jetbrains" : {
+      "backend" : "RubyMine"
```
kou: Could you align style?

```suggestion
  "customizations": {
    "jetbrains": {
      "backend": "RubyMine"
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+    }
+  },
```
kou:
```suggestion
  }
```

.devcontainer/devcontainer.json
kou, line 1: Can we move this to `.devcontainer/2.5/devcontainer.json` or something?
We may add more devcontainrs for old Rubies.

kou (commented):
(I'm not familiar with devcontainer...)

.devcontainer/devcontainer.json
pboling: On GCC v15 systems (like latest Fedora) installing Ruby < 2.7 is not possible (or I have not figured out how yet).  Eventually all systems will upgrade to GCC v15, so this will be a universal problem soon.

So if we need to debug issues, or validate specs, on the oldest supported Ruby 2.5, a container can help.  I am adding them to all my gems.

pboling: > can we use default RubyGems?

We could but IMO it is ideal to develop against the latest available to avoid potential bugs in gem / bundler.

> can we run bundle install?

Yes, I am somewhat indifferent there.  The reason I did not add `bundle install` is because that is where the build was breaking on Ruby 2.5, until I had all the necessary `apt` packages installed.  So I wanted the container to launch, so I could debug the package defficiency.

.devcontainer/devcontainer.json
pboling: It very well could, but I'm not sure which they are, especially after upgrading to not-slim.  The packages can be trimmed later too.  It was my goal that this be the minimal set to install Ruby, but I may have collected too many.

.devcontainer/devcontainer.json
pboling: Yes, but as you are new to devcontainers you should be aware that this is not vanilla JSON.  Devcontainers use a newer spec called "JSON with comments", and known as `JSONC` in linters.

- https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers#editing-the-devcontainerjson-file

.devcontainer/devcontainer.json
pboling: Yes, that is totally a thing

> If you want to have a choice of dev container configurations in your repository, any alternatives to the .devcontainer/devcontainer.json (or .devcontainer.json) file must be located in their own subdirectory at the path .devcontainer/SUBDIRECTORY/devcontainer.json

- https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers#devcontainerjson

.devcontainer/devcontainer.json
pboling: Oh, and the reason there is a Java package is I am trying to additionally test everything against JRuby going forward.

.devcontainer/devcontainer.json
pboling: Added `bundle install` in 5b7cd87

.devcontainer/devcontainer.json
pboling: Moved to `2.5` subdir in 5b7cd87

pboling:
@kou Ready for review again.

.devcontainer/devcontainer.json
kou: I think that all packages are needless because:
* We don't need to build CRuby (`ruby:2.5-buster` already has built CRuby)
* We don't need JRuby (We should use separated `jruby` image instead)

BTW, how can I test the current configuration on local...? I'm not a Visual Studio Code user. So I want to test without Visual Studio Code...

.devcontainer/devcontainer.json
kou: Oh!

.devcontainer/devcontainer.json
pboling: I have only done it from within RubyMine (in which case you just right click on the file to start the container with a local RubyMine inside the container and a cross mounted project).

But this is how I would do it from the CLi:

https://containers.dev/supporting#devcontainer-cli

pboling: I will remove the apt install script and see if that works.

pboling: I've modified the feature to setup bundler instead of installing apt packages.
```
root@d5a0bdd30fe3:/IdeaProjects/rexml# bundle -v
Bundler version 2.3.27
root@d5a0bdd30fe3:/IdeaProjects/rexml# bundle exec rake
/usr/local/bin/ruby test/run.rb
Loaded suite test
Started
Finished in 4.109169655 seconds.
-----------------------------------------------------------------------------------------------------------
716 tests, 2170 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-----------------------------------------------------------------------------------------------------------
174.24 tests/s, 528.09 assertions/s
```

pboling: In doing so I noticed that this gem has bundler as a dependency in the Gemfile, and that's not a good idea, unless the gem is literally something that hacks on bundler.  This gem only uses bundler in the normal, vanilla, way, so it should not have bundler, even as a development dependency. Added as #276 

@kou This PR is ready for review.

pboling:
@kou this is ready for review.  The Gemfile for Ruby 2.5 is now copied fresh each time the devcontainer is run, and is gitignored.

PR #275, opened 2025-07-09
Fix & Deprecate REXML::Text#text_indent

### LINKED ISSUE #273: REXML::Text#indent_text broken since v3.2.6
Version 3.2.6 of this gem released a major performance improvement from #94 which was largely the result of freezing strings by default, by adding the magoc comment to many files:

```ruby
# frozen_string_literal: true
```

The following method from the `REXML::Text` class has never worked since that time, because it modifies frozen  strings in multiple places.
```
    def indent_text(string, level=1, style="\t", indentfirstline=true)
      return string if level < 0
      new_string = ''
      string.each_line { |line|
        indent_string = style * level
        new_line = (indent_string + line).sub(/[\s]+$/,'')
        new_string << new_line
      }
      new_string.strip! unless indentfirstline
      return new_string
    end
```

I did not catch this myself, rather my IDE (RubyMine) noticed it and redlined it for me.

The fact this has never been reported (until now) is likely the result of a few factors:

1. It is an instance method that is unrelated to the instance.  There is never a reason to call this method on an instance of Text, and it doesn't use Text.  Instead it operates on a string passed in as an argument.
2. The method is never called by any code within the gem.
3. There are no tests for the method.
4. There is a different `indent_text` method in the `Pretty` formatter which probably gets used, and is called from internal code.

As a result I propose we do one of the following:

- Deprecate `REXML::Text#indent_text`, but leave it broken, resulting in throwing an error, as it has since v3.2.6. Remove in v4.
- Deprecate `REXML::Text#indent_text`, but fix it. Remove in v4.

pboling:
- Fixes #273 
  - "Fix" in the sense that it restores the original behavior pre-v3.2.6, regardless of its fitness for purpose.
- [x] Regression Test Added

Replaces #274

### DIFF
--- a/lib/rexml/text.rb
+++ b/lib/rexml/text.rb
@@ -269,7 +269,7 @@ module REXML
     end
 
     def indent_text(string, level=1, style="\t", indentfirstline=true)
-      Kernel.warn("#{self.class.name}#indent_text is deprecated.  See REXML::Formatters", uplevel: 1)
+      Kernel.warn("#{self.class.name}#indent_text is deprecated. See REXML::Formatters", uplevel: 1)
       return string if level < 0
       new_string = ''
       string.each_line { |line|
--- a/lib/rexml/text.rb
+++ b/lib/rexml/text.rb
@@ -278,7 +278,7 @@ module REXML
         new_line = (indent_string + line).sub(/[\s]+$/,'')
         new_string << new_line
       }
-      new_string.strip! unless indentfirstline # no change
+      new_string.strip! unless indentfirstline
       new_string
     end
 
--- a/lib/rexml/text.rb
+++ b/lib/rexml/text.rb
@@ -269,14 +269,15 @@ module REXML
     end
 
     def indent_text(string, level=1, style="\t", indentfirstline=true)
+      Kernel.warn("#{self.class.name}#indent_text is deprecated.  See REXML::Formatters", uplevel: 1)
       return string if level < 0
       new_string = ''
       string.each_line { |line|
         indent_string = style * level
         new_line = (indent_string + line).sub(/[\s]+$/,'')
-        new_string << new_line
+        new_string += new_line
       }
-      new_string.strip! unless indentfirstline
+      new_string = new_string.strip unless indentfirstline
       new_string
     end
 
--- a/lib/rexml/text.rb
+++ b/lib/rexml/text.rb
@@ -271,13 +271,14 @@ module REXML
     def indent_text(string, level=1, style="\t", indentfirstline=true)
       Kernel.warn("#{self.class.name}#indent_text is deprecated. See REXML::Formatters", uplevel: 1)
       return string if level < 0
-      new_string = ''
+
+      new_string = +''
       string.each_line { |line|
         indent_string = style * level
         new_line = (indent_string + line).sub(/[\s]+$/,'')
-        new_string += new_line
+        new_string << new_line
       }
-      new_string = new_string.strip unless indentfirstline
+      new_string.strip! unless indentfirstline # no change
       new_string
     end
 
--- a/lib/rexml/xpath.rb
+++ b/lib/rexml/xpath.rb
@@ -64,7 +64,7 @@ module REXML
       parser.namespaces = namespaces
       parser.variables = variables
       path = "*" unless path
-      parser.parse(path,element) || []
+      parser.parse(path,element)
     end
   end
 end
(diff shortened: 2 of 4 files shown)


lib/rexml/text.rb
```diff
@@ -269,14 +269,15 @@ def wrap(string, width, addnewline=false)
     end
 
     def indent_text(string, level=1, style="\t", indentfirstline=true)
+      Kernel.warn("#{self.class.name}#indent_text is deprecated.  See REXML::Formatters", uplevel: 1)
```
kou:
```suggestion
      Kernel.warn("#{self.class.name}#indent_text is deprecated. See REXML::Formatters", uplevel: 1)
```

lib/rexml/text.rb
```diff
@@ -269,14 +269,15 @@ def wrap(string, width, addnewline=false)
...
       return string if level < 0
       new_string = ''
       string.each_line { |line|
         indent_string = style * level
         new_line = (indent_string + line).sub(/[\s]+$/,'')
-        new_string << new_line
+        new_string += new_line
       }
-      new_string.strip! unless indentfirstline
+      new_string = new_string.strip unless indentfirstline
```
kou: How about using `new_string = +''` instead?

test/test_text.rb
```diff
@@ -69,5 +69,10 @@ def test_clone
       assert_equal(text.to_s,
                    text.clone.to_s)
     end
+
+    def test_indent_text
+      text = Text.new("")
+      assert_equal("\tline1\tline2\tline3", text.indent_text("line1\r\nline2\r\nline3\r\n"))
```
kou:
```suggestion
      suppress_warning do
        assert_equal("\tline1\tline2\tline3", text.indent_text("line1\r\nline2\r\nline3\r\n"))
      end
```

lib/rexml/text.rb
pboling: I am not sure how that would be equivalent to strip.  Is the spacing correct there?

`= +''`?

kou: Ah, sorry. I wanted to suggest the following:

```ruby
      new_string = +''
      string.each_line { |line|
        indent_string = style * level
        new_line = (indent_string + line).sub(/[\s]+$/,'')
        new_string << new_line # no change
      }
      new_string.strip! unless indentfirstline # no change
```

We can create a non frozen `String` by `String#@+`.

test/test_text.rb
```diff
@@ -69,5 +69,12 @@ def test_clone
       assert_equal(text.to_s,
                    text.clone.to_s)
     end
+
+    def test_indent_text
+      text = Text.new("")
+      suppress_warning do
```
kou, line 76: Ah, sorry. Could you add `include Helper::Global` to this class?

```diff
diff --git a/test/test_text.rb b/test/test_text.rb
index bae2165..17cdec0 100644
--- a/test/test_text.rb
+++ b/test/test_text.rb
@@ -2,6 +2,7 @@
 
 module REXMLTests
   class TextTester < Test::Unit::TestCase
+    include Helper::Global
     include REXML
 
     def test_new_text_response_whitespace_default
```

lib/rexml/text.rb
pboling: Done in 56f5203

test/test_text.rb
pboling: Done in https://github.com/ruby/rexml/pull/275/commits/e1256056942525dbe6a1eb572979f690e965389f

lib/rexml/text.rb
```diff
@@ -269,14 +269,16 @@ def wrap(string, width, addnewline=false)
...
-      new_string = ''
+
+      new_string = +''
       string.each_line { |line|
         indent_string = style * level
         new_line = (indent_string + line).sub(/[\s]+$/,'')
         new_string << new_line
       }
-      new_string.strip! unless indentfirstline
+      new_string.strip! unless indentfirstline # no change
```
kou:
```suggestion
      new_string.strip! unless indentfirstline
```

lib/rexml/xpath.rb
```diff
@@ -64,7 +64,7 @@ def XPath::match(element, path=nil, namespaces=nil, variables={}, options={})
       parser.namespaces = namespaces
       parser.variables = variables
       path = "*" unless path
-      parser.parse(path,element)
+      parser.parse(path,element) || []
```
kou: Can we revert a needless change?

```suggestion
      parser.parse(path,element)
```

pboling: Oops, I intended to leave this out, but it snuck back in.  Thanks for catching it!

lib/rexml/text.rb
pboling: I am very confused.  I am certain I removed that...

PR #279, opened 2025-07-15
[WIP] 📌 Gemfile.lock should always be in version control

pboling:
- https://bundler.io/guides/creating_gem.html#testing-our-gem
- [x] Adds `Gemfile.lock` to version control
- [ ] Updates CI to use discrete gemfiles for each version of Ruby supported
  - NOTE: These will *not* have their `.lock` files committed to version control.


pboling:
Abandoning, as maintainers have a different view of Gemfile.lock for libraries.  Happy to revisit if that changes.

PR #282, opened 2025-08-11
Added XML declaration check & `Source#skip_spaces` method

naitoh:
### Added XML declaration check

- The version attribute is required in XML declaration.
- Only version attribute, encoding attribute, and standalone attribute are allowed in XML declaration.
- XML declaration is only allowed once.

See: https://www.w3.org/TR/xml/#NT-XMLDecl

### Added `Source#skip_spaces` method

In the case of `@source.match?(/\s+/um, true)`, if there are no spaces at the beginning, I want to stop reading immediately.
However, it continues to read the buffer until it finds a match, but it never finds a match.
As a result, it continues reading until the end of the file.

In the case of large XML files, drop_parsed_content occur frequently until the buffer is cleared, which may affect performance.

## Benchmark

```
                 dom     32.534      35.130        54.559       53.528 i/s -     100.000 times in 3.073715s 2.846540s 1.832883s 1.868189s
                 sax     44.785      44.089        78.303       77.842 i/s -     100.000 times in 2.232907s 2.268138s 1.277093s 1.284657s
                pull     51.750      51.105        90.819       90.658 i/s -     100.000 times in 1.932351s 1.956759s 1.101094s 1.103050s
              stream     51.427      51.444        89.820       88.971 i/s -     100.000 times in 1.944502s 1.943855s 1.113340s 1.123960s

Comparison:
                              dom
        before(YJIT):        54.6 i/s 
         after(YJIT):        53.5 i/s - 1.02x  slower
               after:        35.1 i/s - 1.55x  slower
              before:        32.5 i/s - 1.68x  slower

                              sax
        before(YJIT):        78.3 i/s 
         after(YJIT):        77.8 i/s - 1.01x  slower
              before:        44.8 i/s - 1.75x  slower
               after:        44.1 i/s - 1.78x  slower

                             pull
        before(YJIT):        90.8 i/s 
         after(YJIT):        90.7 i/s - 1.00x  slower
              before:        51.8 i/s - 1.75x  slower
               after:        51.1 i/s - 1.78x  slower

                           stream
        before(YJIT):        89.8 i/s 
         after(YJIT):        89.0 i/s - 1.01x  slower
               after:        51.4 i/s - 1.75x  slower
              before:        51.4 i/s - 1.75x  slower
```

- YJIT=ON : 0.98x - 1.00x faster
- YJIT=OFF : 0.98x - 1.07x faster

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -280,7 +280,7 @@ module REXML
               return [ :comment, process_comment ]
             elsif @source.match?("DOCTYPE", true)
               base_error_message = "Malformed DOCTYPE"
-              unless @source.match?(/\s+/um, true)
+              unless @source.skip_spaces
                 if @source.match?(">")
                   message = "#{base_error_message}: name is missing"
                 else
@@ -290,7 +290,7 @@ module REXML
                 raise REXML::ParseException.new(message, @source)
               end
               name = parse_name(base_error_message)
-              @source.match?(/\s*/um, true) # skip spaces
+              @source.skip_spaces
               if @source.match?("[", true)
                 id = [nil, nil, nil]
                 @document_status = :in_doctype
@@ -306,7 +306,7 @@ module REXML
                   # For backward compatibility
                   id[1], id[2] = id[2], nil
                 end
-                @source.match?(/\s*/um, true) # skip spaces
+                @source.skip_spaces
                 if @source.match?("[", true)
                   @document_status = :in_doctype
                 elsif @source.match?(">", true)
@@ -319,7 +319,7 @@ module REXML
               end
               args = [:start_doctype, name, *id]
               if @document_status == :after_doctype
-                @source.match?(/\s*/um, true)
+                @source.skip_spaces
                 @stack << [ :end_doctype ]
               end
               return args
@@ -330,7 +330,7 @@ module REXML
           end
         end
         if @document_status == :in_doctype
-          @source.match?(/\s*/um, true) # skip spaces
+          @source.skip_spaces
           start_position = @source.position
           if @source.match?("<!", true)
             if @source.match?("ELEMENT", true)
@@ -391,7 +391,7 @@ module REXML
               return [ :attlistdecl, element, pairs, contents ]
             elsif @source.match?("NOTATION", true)
               base_error_message = "Malformed notation declaration"
-              unless @source.match?(/\s+/um, true)
+              unless @source.skip_spaces
                 if @source.match?(">")
                   message = "#{base_error_message}: name is missing"
                 else
@@ -404,7 +404,7 @@ module REXML
               id = parse_id(base_error_message,
                             accept_external_id: true,
                             accept_public_id: true)
-              @source.match?(/\s*/um, true) # skip spaces
+              @source.skip_spaces
               unless @source.match?(">", true)
                 message = "#{base_error_message}: garbage before end >"
                 raise REXML::ParseException.new(message, @source)
@@ -425,7 +425,7 @@ module REXML
           end
         end
         if @document_status == :after_doctype
-          @source.match?(/\s*/um, true)
+          @source.skip_spaces
         end
         begin
           start_position = @source.position
@@ -735,7 +735,7 @@ module REXML
 
       def process_instruction
         name = parse_name("Malformed XML: Invalid processing instruction node")
-        if @source.match?(/\s+/um, true)
+        if @source.skip_spaces
           match_data = @source.match(/(.*?)\?>/um, true)
           unless match_data
             raise ParseException.new("Malformed XML: Unclosed processing instruction", @source)
@@ -817,7 +817,7 @@ module REXML
               message = "Missing attribute value end quote: <#{name}>: <#{quote}>"
               raise REXML::ParseException.new(message, @source)
             end
-            @source.match?(/\s*/um, true)
+            @source.skip_spaces
             if prefix == "xmlns"
               if local_part == "xml"
                 if value != Private::XML_PREFIXED_NAMESPACE
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -144,6 +144,7 @@ module REXML
         PEREFERENCE_PATTERN = /#{PEREFERENCE}/um
         TAG_PATTERN = /((?>#{QNAME_STR}))\s*/um
         CLOSE_PATTERN = /(#{QNAME_STR})\s*>/um
+        EQUAL_PATTERN = /\s*=\s*/um
         ATTLISTDECL_END = /\s+#{NAME}(?:#{ATTDEF})*\s*>/um
         NAME_PATTERN = /#{NAME}/um
         GEDECL_PATTERN = "\\s+#{NAME}\\s+#{ENTITYDEF}\\s*>"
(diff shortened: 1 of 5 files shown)


kou:
> ### IOSource#match? read_keep param added
> 
> In the case of `@source.match?(/\s+/um, true)`, if there are no spaces at the beginning, I want to stop reading immediately. However, it continues to read the buffer until it finds a match, but it never finds a match. As a result, it continues reading until the end of the file.
> 
> In the case of large XML files, drop_parsed_content occur frequently until the buffer is cleared, which may affect performance.

How about adding `Source#skip_spaces` or something instead of adding a new argument to `Source#match?`?
I feel that adding `read_keep` parameter reduces readability...

naitoh:
I modified it to use `Source#skip_spaces?`.

lib/rexml/source.rb
```diff
@@ -142,14 +143,18 @@ def match(pattern, cons=false)
       end
     end
 
-    def match?(pattern, cons=false)
+    def match?(pattern, cons=false, read_keep=true)
```
kou: Can we revert this change?

lib/rexml/source.rb
```diff
@@ -142,14 +143,18 @@ def match(pattern, cons=false)
...
       if cons
         !@scanner.skip(pattern).nil?
       else
         !@scanner.match?(pattern).nil?
       end
     end
 
+    def skip_spaces?
```
kou: We don't need to use the `?` suffix. Because this is not a predicate method. This returns a boolean but this method's main feature is skipping preceding spaces.

lib/rexml/parsers/baseparser.rb
```diff
@@ -735,37 +736,93 @@ def process_comment
...
+
+      def xml_declaration
+        unless @version.nil?
+          raise ParseException.new("Malformed XML: XML declaration is duplicated", @source)
+        end
+        if @document_status
+          raise ParseException.new("Malformed XML: XML declaration is not at the start", @source)
+        end
+        unless @source.skip_spaces?
+          raise ParseException.new("Malformed XML: XML declaration is missing spaces", @source)
```
kou:
```suggestion
          raise ParseException.new("Malformed XML: XML declaration misses spaces before version", @source)
```

lib/rexml/parsers/baseparser.rb
```diff
@@ -735,37 +736,93 @@ def process_comment
...
+        end
+        unless @source.match?("version", true)
+          raise ParseException.new("Malformed XML: XML declaration is missing version", @source)
```
kou:
```suggestion
          raise ParseException.new("Malformed XML: XML declaration misses version", @source)
```

lib/rexml/parsers/baseparser.rb
```diff
@@ -735,37 +736,93 @@ def process_comment
...
+        unless @source.skip_spaces?
           unless @source.match?("?>", true)
-            raise ParseException.new("Malformed XML: Unclosed processing instruction", @source)
+            raise ParseException.new("Malformed XML: Unclosed XML declaration", @source)
           end
-        end
-        if name == "xml"
-          if @document_status
-            raise ParseException.new("Malformed XML: XML declaration is not at the start", @source)
+          if /\AUTF-16(?:BE|LE)\z/i =~ @source.encoding
```
kou:
```suggestion
          if /\AUTF-16(?:BE|LE)\z/i.match?(@source.encoding)
```

lib/rexml/parsers/baseparser.rb
```diff
@@ -735,37 +736,93 @@ def process_comment
...
+          encoding = get_attribute("xml")
+          unless @source.skip_spaces?
+            unless @source.match?("?>", true)
+              raise ParseException.new("Malformed XML: Unclosed XML declaration", @source)
+            end
+            if need_source_encoding_update?(encoding)
+              @source.encoding = encoding
+            end
+
+            if encoding.nil? and /\AUTF-16(?:BE|LE)\z/i =~ @source.encoding
```
kou:
```suggestion
            if encoding.nil? and /\AUTF-16(?:BE|LE)\z/i.match?(@source.encoding)
```

BTW, can we unify these `UTF-16` normalization process?

lib/rexml/parsers/baseparser.rb
```diff
@@ -787,6 +844,25 @@ def scan_quote
         end
       end
 
+      def get_attribute(name)
```
kou: How about `parse_attribute_value_with_equal` or something?

lib/rexml/source.rb
naitoh: Oh, sorry.

lib/rexml/source.rb
naitoh: OK.
I see.

lib/rexml/parsers/baseparser.rb
naitoh: OK.
I see.

lib/rexml/parsers/baseparser.rb
naitoh: OK.
I see.

lib/rexml/parsers/baseparser.rb
naitoh: OK.
I see.

lib/rexml/parsers/baseparser.rb
naitoh: OK.
I see.

lib/rexml/parsers/baseparser.rb
naitoh: OK.
I see.

lib/rexml/parsers/baseparser.rb
```diff
@@ -735,37 +743,93 @@ def process_comment
...
-            raise ParseException.new("Malformed XML: Unclosed processing instruction", @source)
+            raise ParseException.new("Malformed XML: Unclosed XML declaration", @source)
           end
-        end
-        if name == "xml"
-          if @document_status
-            raise ParseException.new("Malformed XML: XML declaration is not at the start", @source)
+          if utf16_variant_encoding?(@source.encoding)
+            encoding = "UTF-16"
           end
```
kou: How about defining `normalize_xml_declaration_encoding` that can be used like `encoding = normalize_xml_declaration_encoding(@source.encoding)`?

lib/rexml/parsers/baseparser.rb
```diff
@@ -290,7 +293,7 @@ def pull_event
                 raise REXML::ParseException.new(message, @source)
               end
               name = parse_name(base_error_message)
-              @source.match?(/\s*/um, true) # skip spaces
+              @source.match?(Private::CUT_SPACES, true)
```
kou: Can we use `@source.skip_spaces` here too?

naitoh: Oh, it is cool!
Thanks!

lib/rexml/source.rb
```diff
@@ -65,9 +65,10 @@ class Source
     attr_reader :encoding
 
     module Private
+      SKIP_SPACES = /\s+/um
```
tompng: About naming. How about `SPACES` or `SPACES_PATTERN`?
I think most regexp constant names in REXML represents what it is or what it matches, but unlike other constants, the naming of `SKIP_SPACES` and `CUT_SPACES` focuses on action.

If it is only used in `def skip_spaces`, simply `@scanner.skip(/\s+/um)` might be good enough.

naitoh: I changed to `SPACES_PATTERN`.
Thank you!

lib/rexml/parsers/baseparser.rb
naitoh: OK, I see.
Thanks!

kou (approved):
+1

naitoh:
Thank you for your reviews.

PR #286, opened 2025-08-26
Fix ParseError when String has no trailing newline character

leoarnold:


kou:
Thanks but we should reject no root XML because it's an invalid XML. So we need a different approach.

PR #288, opened 2025-08-28
Fixed initialization processing when reaching the end of a file in `IOSource#read`

naitoh:
## Why?
Avoid causing an error when `@source.eof?` is called after `IOSource#read_until` is invoked in subsequent processing.

> /.gems/gems/rexml-3.4.2/lib/rexml/parsers/baseparser.rb:524:in 'REXML::Parsers::BaseParser#pull_event': #<NoMethodError: undefined method 'eof?' for nil> (REXML::ParseException)
> .gems/gems/rexml-3.4.2/lib/rexml/source.rb:275:in 'REXML::IOSource#read_until'

https://github.com/ruby/rexml/blob/185bdc737da406ba4f9564726849ad3477858eb2/lib/rexml/source.rb#L266-L276

See: https://github.com/ruby/rexml/issues/287


lib/rexml/source.rb
```diff
@@ -257,9 +257,11 @@ def read(term = nil, min_bytes = 1)
         rescue IOError
         end
         true
-      rescue Exception, NameError
+      rescue ArgumentError
         @source = nil
         false
+      rescue Exception, NameError
+        false
```
kou: Hmm. If we don't set `nil` to `@source`, `@source` isn't GC-ed, right?

Can we choose another approach something like the following?

```diff
diff --git a/lib/rexml/source.rb b/lib/rexml/source.rb
index 9950007..8b8ba0d 100644
--- a/lib/rexml/source.rb
+++ b/lib/rexml/source.rb
@@ -272,7 +272,7 @@ module REXML
         @scanner << readline(term)
       end
       if str
-        read if @scanner.eos? and !@source.eof?
+        read if @scanner.eos? and @source and !@source.eof?
         str
       else
         rest = @scanner.rest
```

BTW, can we add a test for this case like #286 does?

naitoh: > Can we choose another approach something like the following?

OK, I see.

> BTW, can we add a test for this case like https://github.com/ruby/rexml/pull/286 does?

OK, I see.

test/test_io_source.rb
```diff
@@ -0,0 +1,18 @@
+require "rexml/source"
+
+module REXMLTests
+  class TestIOSource < Test::Unit::TestCase
+    include REXML
+    def setup
+      @xml = SourceFactory.create_from '<?xml version="1.0"?>'
```
kou:
```suggestion
      @source = SourceFactory.create_from('<?xml version="1.0"?>')
```

kou: Could you avoid using `include` as much as possible to avoid naming conflict?

```suggestion
    def setup
      @xml = REXML::SourceFactory.create_from '<?xml version="1.0"?>'
```

test/test_io_source.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+
+    sub_test_case("#read_until") do
+      test("eof") do
+        @xml.read(">")
+        @xml.read(">") # Set EOF
```
kou:
```suggestion
        assert_true(@xml.read("nonexistent")) # Consume all data
        assert_false(@xml.read("nonexistent")) # Set EOF
```

test/test_io_source.rb
naitoh: OK, I see.

naitoh: OK, I see.

test/test_io_source.rb
naitoh: OK, I see.

PR #293, opened 2025-09-07
Add 3.4.3 entry

naitoh:


NEWS.md
```diff
@@ -1,5 +1,35 @@
...
+
+  * Reject no root element XML as an invalid XML
+    * GH-289
+    * GH-291
+    * Patch by NAITOH Jun
+    * Reported by Sutou Kouhei
+
+### Fixes
+
+  * Fixed an issue with IOSource#read_until when reaching the end of a file
```
kou:
```suggestion
  * Fixed an issue with `IOSource#read_until` when reaching the end of a file
```

NEWS.md
```diff
@@ -1,5 +1,35 @@
...
+    * GH-287
+    * GH-288
+    * Patch by NAITOH Jun
+    * Reported by Jason Thomas
+
+  * The Zlib::GzipReader in JRuby does not behave as expected with REXML, so the test is skipped
```
kou:
```suggestion
  * The `Zlib::GzipReader` in JRuby does not behave as expected with REXML, so the test is skipped
```

NEWS.md
```diff
@@ -1,5 +1,35 @@
...
+    * GH-292
+    * Patch by NAITOH Jun
```
kou: We can omit this because this changes only tests.

NEWS.md
naitoh: OK, I see.

NEWS.md
naitoh: OK, I see.

NEWS.md
naitoh: OK, I see.

PR #291, opened 2025-09-06
Reject no root element XML as an invalid XML

naitoh:
## Why?

GitHub: fix https://github.com/ruby/rexml/issues/289

We must reject all well-formed XMLs:

https://www.w3.org/TR/2006/REC-xml11-20060816/#proc-types

> Validating and non-validating processors alike MUST report violations of this specification's well-formedness constraints in the content of the [document entity](https://www.w3.org/TR/2006/REC-xml11-20060816/#dt-docent) and any other [parsed entities](https://www.w3.org/TR/2006/REC-xml11-20060816/#dt-parsedent) that they read.

No root element XML is not well-formed because `document` requires one `element`:

https://www.w3.org/TR/2006/REC-xml11-20060816/#sec-well-formed

> [1]   	document	   ::=   	( [prolog](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-prolog) [element](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-element) [Misc](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-Misc)* ) - ( [Char](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-Char)* [RestrictedChar](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-RestrictedChar) [Char](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-Char)* )

PR #295, opened 2025-09-09
Change `REXML::Document.new("")` to behave the same as `REXML::Document.new(nil)`

naitoh:
## Why?

XML without a root element is invalid XML, but to maintain compatibility, we will change the behavior of `REXML::Document.new("")` to be the same as `REXML::Document.new(nil)`.

See: https://github.com/ruby/rexml/pull/291#issuecomment-3266569157


test/test_core.rb
```diff
@@ -1550,6 +1550,10 @@ def test_ticket_138
                    REXML::Document.new(doc.root.to_s).root.attributes.to_h)
     end
 
+    def test_empty_doc
+      assert(REXML::Document.new('').children.empty?)
```
kou: * Could you use power assert for better failure message?
* Could you also add a test for `REXML::Document.new` if it doesn't exist?
* Could you add a comment that this must be rejected but we accept it for backward compatibility?

```suggestion
      # This is an invalid usage because no root element XML is an invalid XML.
      # But we accept `""` for backward compatibility.
      assert do
        REXML::Document.new('').children.empty?
      end
```

lib/rexml/document.rb
```diff
@@ -95,7 +95,7 @@ def initialize( source = nil, context = {} )
       @entity_expansion_text_limit = Security.entity_expansion_text_limit
       super()
       @context = context
-      return if source.nil?
+      return if source.nil? or source == ""
```
kou, line 100: Could you add a comment why we accept `""`?

test/test_core.rb
naitoh: OK, I see.
Done!

lib/rexml/document.rb
naitoh: OK, I see.
Done!

test/test_core.rb
```diff
@@ -1550,6 +1550,22 @@ def test_ticket_138
                    REXML::Document.new(doc.root.to_s).root.attributes.to_h)
     end
 
+    def test_document_with_no_arguments
+      assert(REXML::Document.new.children.empty?)
```
kou: Could you use power assert?

test/test_core.rb
```diff
@@ -1550,6 +1550,22 @@ def test_ticket_138
...
+    end
+
+    def test_document_with_nil_arguments
+      assert(REXML::Document.new(nil).children.empty?)
```
kou: ditto.

test/test_core.rb
naitoh: Oh, sorry.

test/test_core.rb
naitoh: Oh, sorry.

lib/rexml/document.rb
```diff
@@ -95,7 +95,9 @@ def initialize( source = nil, context = {} )
       @entity_expansion_text_limit = Security.entity_expansion_text_limit
       super()
       @context = context
-      return if source.nil?
+      # This is an invalid usage because no root element XML is an invalid XML.
```
kou:
```suggestion
      # `source = ""` is an invalid usage because no root element XML is an invalid XML.
```

PR #299, opened 2025-09-13
Use more StringScanner based API for parse_id

naitoh:
## Why?

Improve maintainability by optimizing the process so that the parsing process proceeds using StringScanner.

### DIFF
--- a/lib/rexml/parsers/baseparser.rb
+++ b/lib/rexml/parsers/baseparser.rb
@@ -310,7 +310,6 @@ module REXML
                 @source.ensure_buffer
               else
                 id = parse_id(base_error_message,
-                              accept_external_id: true,
                               accept_public_id: false)
                 if id[0] == "SYSTEM"
                   # For backward compatibility
@@ -412,7 +411,6 @@ module REXML
               end
               name = parse_name(base_error_message)
               id = parse_id(base_error_message,
-                            accept_external_id: true,
                             accept_public_id: true)
               @source.skip_spaces
               unless @source.match?(">", true)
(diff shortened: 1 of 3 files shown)


lib/rexml/parsers/baseparser.rb
```diff
@@ -667,69 +668,46 @@ def parse_name(base_error_message)
       end
 
       def parse_id(base_error_message,
-                   accept_external_id:,
```
kou, line 670: We can remove this keyword argument because it's always `true`, right?

lib/rexml/parsers/baseparser.rb
```diff
@@ -667,69 +668,46 @@ def parse_name(base_error_message)
...
-          if accept_public_id
-            if @source.match?(/#{public}\s+#{PUBIDLITERAL}\s+[^'"]/um)
-              return "system ID literal is missing"
-            end
-            unless @source.match?(/#{public}\s+#{PUBIDLITERAL}\s+#{SYSTEMLITERAL}/um)
-              return "invalid system literal"
-            end
-            "garbage after system literal"
+          if @source.match?(/(?:\s+[^'"]|\s*[\[>])/um)
+            details = "public ID literal is missing"
```
kou: How about raising here for better backtrace?

lib/rexml/parsers/baseparser.rb
naitoh: Yes.

https://github.com/ruby/rexml/blob/39e7e28f79705d8e50513330f26e67195bece819/lib/rexml/parsers/baseparser.rb#L309-L310

https://github.com/ruby/rexml/blob/39e7e28f79705d8e50513330f26e67195bece819/lib/rexml/parsers/baseparser.rb#L411-L412

lib/rexml/parsers/baseparser.rb
naitoh: OK, I see.

PR #300, opened 2025-09-15
Optimization of REXML::Parsers::SAX2Parser#get_namespace

naitoh:
## Benchmark
```
$ benchmark-driver benchmark/sax.yaml
                 sax     1.362k      2.207k        1.713k       2.751k i/s -     100.000 times in 0.073410s 0.045312s 0.058386s 0.036345s

Comparison:
                              sax
         after(YJIT):      2751.4 i/s
               after:      2206.9 i/s - 1.25x  slower
        before(YJIT):      1712.7 i/s - 1.61x  slower
              before:      1362.2 i/s - 2.02x  slower
```

- YJIT=ON : 1.61x faster
- YJIT=OFF : 1.62x faster


lib/rexml/parsers/sax2parser.rb
```diff
@@ -122,7 +122,7 @@ def parse
               event[2].each {|n, v| event[2][n] = @parser.normalize(v)}
               nsdecl = event[2].find_all { |n, value| n =~ /^xmlns(:|$)/ }
               nsdecl.collect! { |n, value| [ n[6..-1], value ] }
-              @namespace_stack.push({})
+              @namespace_stack.push(@namespace_stack.empty? ? {} : @namespace_stack[-1].dup)
```
kou: How about initializing `@namespace_stack` as `[{}]` to simplify implementation?

lib/rexml/parsers/sax2parser.rb
```diff
@@ -260,10 +260,7 @@ def add( pair )
 
       def get_namespace( prefix )
         return nil if @namespace_stack.empty?
-
-        uris = (@namespace_stack.find_all { |ns| not ns[prefix].nil? }) ||
-          (@namespace_stack.find { |ns| not ns[nil].nil? })
-        uris[-1][prefix] unless uris.nil? or 0 == uris.size
+        @namespace_stack[-1][prefix] || @namespace_stack[-1][nil]
```
kou: We can avoid multiple `namespace_stack[-1]` calls here:

```suggestion
        current_namespace = @namespace_stack[-1]
        current_namespace[prefix] || current_namespace[nil]
```

naitoh: OK, I see.

lib/rexml/parsers/sax2parser.rb
naitoh: OK, I see.

lib/rexml/parsers/sax2parser.rb
```diff
@@ -260,10 +260,8 @@ def add( pair )
 
       def get_namespace( prefix )
         return nil if @namespace_stack.empty?
```
kou: Can we remove this now?

naitoh: OK, I see.

PR #307, opened 2025-12-08
[DOC] Omit include files from html

BurdetteLamar:
Five pages listed in the left pane at https://ruby.github.io/rexml/ are in fact intended only as include files, and not as independent pages;  these files have links that work fine when the files are included (as I intended), but are broken when rendered as separate html (which was not what I intended).

These pages should not be included in the html.  The pages are in directory `doc/`:

- [child_toc](https://ruby.github.io/rexml/doc/rexml/tasks/tocs/child_toc_rdoc.html)
- [document_toc](https://ruby.github.io/rexml/doc/rexml/tasks/tocs/document_toc_rdoc.html)
- [element_toc](https://ruby.github.io/rexml/doc/rexml/tasks/tocs/element_toc_rdoc.html)
- [node_toc](https://ruby.github.io/rexml/doc/rexml/tasks/tocs/node_toc_rdoc.html)
- [parent_toc](https://ruby.github.io/rexml/doc/rexml/tasks/tocs/parent_toc_rdoc.html)

The [master_toc](https://ruby.github.io/rexml/doc/rexml/tasks/tocs/master_toc_rdoc.html) is okay, and should be included in the html;  other pages link to it.

This PR moves those files to a new directory, `includes/`, which is omitted from the html;  the remaining files stay in place, but their directory is renamed to `rdoc/`.

Also fixes links to tasks from Document and Element class pages.


kou:
What is fixed by this? Could you explain it in the PR description?

We don't want to list `*_toc` in "Pages" -> "doc" in the sidebar in https://ruby.github.io/rexml/ ?

image

BurdetteLamar:
> What is fixed by this? Could you explain it in the PR description?
> 

Explanation added to PR description.

lib/rexml/document.rb
```diff
@@ -28,9 +28,9 @@ module REXML
   # If you're somewhat familiar with XML
   # and have a particular task in mind,
   # you may want to see the
-  # {tasks pages}[../doc/rexml/tasks/tocs/master_toc_rdoc.html],
+  # {tasks pages}[../doc/rexml/tasks/rdoc/master_toc_rdoc.html],
```
kou: Should we update this path and other similar paths?


```suggestion
  # {tasks pages}[../doc/rexml/tasks/rdoc/toc_rdoc.html],
```

BurdetteLamar: Done.

kou:
Thanks.

PR #310, opened 2026-05-05
Optimization of REXML::XPathParser#step

naitoh:
## Benchmark

```
  REXML::XPath.match(REXML::Document.new(xml), 'a//a')     4.170k      4.300k        4.091k       4.087k i/s -     100.000 times in 0.023979s 0.023257s 0.024444s 0.024470s
REXML::XPath.match(REXML::Document.new(xml), '//a//a')    196.250      1.378k       293.984       1.701k i/s -     100.000 times in 0.509554s 0.072574s 0.340154s 0.058796s

Comparison:
                                                 after:      4299.8 i/s 
                                                before:      4170.3 i/s - 1.03x  slower
                                          before(YJIT):      4091.0 i/s - 1.05x  slower
                                           after(YJIT):      4086.6 i/s - 1.05x  slower

             REXML::XPath.match(REXML::Document.new(xml), '//a//a')
                                           after(YJIT):      1700.8 i/s 
                                                 after:      1377.9 i/s - 1.23x  slower
                                          before(YJIT):       294.0 i/s - 5.79x  slower
                                                before:       196.3 i/s - 8.67x  slower
```
- YJIT=ON : 0.99x - 5.79x faster
- YJIT=OFF : 1.03x - 7.01x faster

### DIFF
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -462,21 +462,20 @@ module REXML
         if nodesets.size == 1
           ordered_nodeset = nodesets[0]
         else
+          seen = {}.compare_by_identity
           raw_nodes = []
           nodesets.each do |nodeset|
             nodeset.each do |node|
-              if node.respond_to?(:raw_node)
-                raw_nodes << node.raw_node
-              else
-                raw_nodes << node
-              end
+              raw_node = node.respond_to?(:raw_node) ? node.raw_node : node
+              next if seen.key?(raw_node)
+              seen[raw_node] = true
+              raw_nodes << raw_node
             end
           end
           ordered_nodeset = sort(raw_nodes, order)
         end
         new_nodeset = []
         ordered_nodeset.each do |node|
-          # TODO: Remove duplicated
           new_nodeset << XPathNode.new(node, position: new_nodeset.size + 1)
         end
         new_nodeset
--- a/test/xpath/test_base.rb
+++ b/test/xpath/test_base.rb
@@ -1306,5 +1306,79 @@ EOF
       result = XPath.match(xmldoc, "//a//a")
       assert_equal(["1", "2", "3"], result.map { |e| e.attributes["id"] })
     end
+
+    def test_descendant_ancestor_descendant_overlap
+      doc = <<-XML
+      <root>
+        <a id='1'>
+          <a id='2'>
+            <b id='b1'/>
+          </a>
+        </a>
+      </root>
+      XML
+
+      xmldoc = Document.new(doc)
+      result = XPath.match(xmldoc, "//a//b")
+      assert_equal(["b1"], result.map { |e| e.attributes["id"] })
+    end
+
+    def test_descendant_sibling_branches
+      doc = <<-XML
+      <root>
+        <a id='1'>
+          <b id='b1'/>
+        </a>
+        <a id='2'>
+          <b id='b2'/>
+        </a>
+      </root>
+      XML
+      xmldoc = Document.new(doc)
+      result = XPath.match(xmldoc, "//a//b")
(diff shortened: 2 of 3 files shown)


lib/rexml/xpath_parser.rb
```diff
@@ -685,14 +684,12 @@ def sort(array_of_nodes, order)
     end
 
     def descendant(nodeset, include_self)
-      nodesets = []
+      new_nodeset = []
+      new_nodes = {}
       nodeset.each do |node|
-        new_nodeset = []
-        new_nodes = {}
```
tompng, line 691: With this change, the result of this xpath has changed.
```ruby
REXML::XPath.match(REXML::Document.new('<a id="a1"><b id="b1"><a id="a2"><b id="b2"><b id="b3"></b></b></a></b></a>'), '//a/descendant::b[1]')
# => [<b id='b1'> ... </>, <b id='b2'> ... </>] Expected (in master branch)
# => [<b id='b1'> ... </>] Actual (this pull request)
```

Reference:
```js
document.body.innerHTML='<a id="a1"><b id="b1"><a id="a2"><b id="b2"><b id="b3"></b></b></a></b></a>'
snapshot = document.evaluate('//a/descendant::b[1]', document, null, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE, null);
[...new Array(snapshot.snapshotLength)].map((_, i) => snapshot.snapshotItem(i))
// => [b#b1, b#b2]
```

lib/rexml/xpath_parser.rb
```diff
@@ -462,21 +462,20 @@ def step(path_stack, any_type: :element, order: :forward)
...
           nodesets.each do |nodeset|
             nodeset.each do |node|
-              if node.respond_to?(:raw_node)
-                raw_nodes << node.raw_node
-              else
-                raw_nodes << node
-              end
+              raw_node = node.respond_to?(:raw_node) ? node.raw_node : node
+              next if seen.key?(raw_node.object_id)
+              seen[raw_node.object_id] = true
```
tompng: `object_id` will lazily assign a new id when it is called. We can avoid this by using `seen = {}.compare_by_identity` instead.

lib/rexml/xpath_parser.rb
naitoh: I'm sorry.
Since this change has unintended consequences, I will cancel the change to #descendant.
Thanks!

lib/rexml/xpath_parser.rb
naitoh: Thanks for pointing that out.
I’ll fix it.

tompng: Using `compare_by_identity`, Hash will compare by its identity, (same as comparing by object_id, but without assigning actual object_id), so `seen.key?(raw_node)` and `seen[raw_node] = true` is enough.

These two code gives the same result
```ruby
h = {}
h[key.object_id] = value
h[key.object_id] #=> value
h[key.dup.object_id] #=> nil
```
```ruby
h = {}.compare_by_identity
h[key] = value
h[key] #=> value
h[key.dup] #=> nil
```

naitoh: OK, I see.
Fixed it.

test/xpath/test_base.rb
```diff
@@ -1276,5 +1276,109 @@ def test_match_with_deprecated_usage
     ensure
       $VERBOSE = verbose
     end
+
+    def test_descendant_no_duplicates
+      doc = <<-EOT
```
kou: Could you use `XML` not `EO...` for here document mark for readability?

kou (approved):
+1

test/xpath/test_base.rb
naitoh: OK.
I see.

kou:
Thanks.

PR #315, opened 2026-05-21
Optimize XPath step

tompng:
Refactor `step` so that nodeset materialization is deferred. Instead of building the full nodeset up front and filtering through predicates,
each axis returns a *scan descriptor* (`[generator_name, generator_argument]`), and `step` picks a scan strategy based on the predicates' shape.

Predicates are classified into three groups:

| kind | examples | strategy passed to the generator |
|---|---|---|
| position-independent | `[@a="1"]`, `[name()="foo"]`, `[@a=@b]` | `:uniq` — emit deduplicated matching nodes |
| simple positional | `[N]`, `[position()=N]`, `[position()>N]`, `[position()<N]` | `[op, value]` — positional scan with one comparison |
| complex / position-dependent | `[position()*@a]`, `[last()-1]`, ... | `:nodesets` — fall back to per-anchor nodesets + the previous `evaluate_predicate` pipeline |

Mixed predicate lists are split: position-independent predicates *before* the first positional predicate are folded into the node test;
predicates *after* it are applied per-node on the result.

Each axis can implement zero, one, or all of the three strategies. If a strategy is not implemented, the generator falls back to producing
`:nodesets` and the common slow path (`non_optimized_nodesets_select`) handles dedup / positional filtering on flattened nodesets — i.e. the
same behavior as before this PR.

This pull request adds fast paths for:

- `descendant` / `descendant-or-self`: `:uniq` (single DFS with a seen-set; this is what speeds up `//a//a//a//a`)
- `ancestor` / `ancestor-or-self`: `:uniq` (parent-chain walk with a seen-set)
- `preceding-sibling` / `following-sibling`: `:uniq` and `[op, value]` (sibling scan with anchor-index tracking)

Other axes (`child`, `parent`, `self`, `attribute`, `preceding`, `following`, etc) keeps the previous behavior via the fallback path; they can be optimized in follow-ups without changing call sites.

## Detail

For `//a//a//a` style queries, the previous code built nodesets keyed by each anchor, including the same descendant once per anchor. The new
`:uniq` path scans every node at most once per step.

For `[position() > N]` style predicates on wide trees (e.g. `//a/preceding-sibling::*[position()>2]`), we previously built the full
preceding-sibling nodeset for each anchor and then ran `evaluate_predicate`. The new `[op, value]` path scans children once per parent and uses
anchor-index bookkeeping to recover per-anchor positions.

Note: general XPath cannot be linear — e.g. `*[position() * number(@a) % number(@b) = 1]` is genuinely O(n²) — so the goal is only to add a fast-path for specific case: position-independent predicates and simple-positional predicates.

## Benchmark of best case

```ruby
DEPTH = 500
xml   = '<a>' * DEPTH + '</a>' * DEPTH
doc   = REXML::Document.new(xml)
WIDTH = 1000
xml_wide = '<root>' + '<child/>' * WIDTH + '</root>'
doc_wide = REXML::Document.new(xml_wide)

REXML::XPath.match(doc, "//a//a");
# processing time: 30.756939s → 0.126807s

REXML::XPath.match(doc_wide, "//*/preceding-sibling::*[position()=10]");
# processing time: 2.446333s → 0.083954s
```

### Scenario
```yaml
prelude: |
  require "rexml"
  xml_wide = "<root>" + (1..1000).map { |i| "<item id='#{i}'/>" }.join + "</root>"
  wide = REXML::Document.new(xml_wide)
  xml_deep = "<root>" + (1..1000).map { |i| "<item id='#{i}'>" }.join + '</item>'*1000 + "</root>"
  deep = REXML::Document.new(xml_deep)

benchmark:
  child: REXML::XPath.match(wide, "root/item")
  descendant: REXML::XPath.match(deep, "//item")
  descendant-descendant: REXML::XPath.match(deep, "//item//item")
  descendant-descendant-wildcard: REXML::XPath.match(deep, "//*//*")
  ancestor-descendant: REXML::XPath.match(deep, "descendant::*/ancestor::*/descendant::*")
  preceding-following-sibling: REXML::XPath.match(wide, "//*/preceding-sibling::*/following-sibling::*")
  preceding-following-sibling-positional: REXML::XPath.match(wide, "//*/preceding-sibling::*[10]/following-sibling::*[10]")
```

### Compares
master, xpath_step_optimize (this pull), sort_on_demand(#330), sort_improve(Emulate ideal sort computation time), and its combinations.

There's no implementation of `sort_improve` yet, so I used the code below to emulate the computational cost of ideal sort.
```ruby
def sort(array_of_nodes)
  # Just spend time to emulate the ideal computational cost of sorting nodes
  parents = Set.new.compare_by_identity
  array_of_nodes.each { parents << it.parent if it.parent }
  4.times do
    # find the common ancestor
    nodes = array_of_nodes
    seen = Set.new.compare_by_identity
    while nodes.size >= 2
      new_nodes = Set.new.compare_by_identity
      nodes.map(&:parent).each do |parent|
        if parent && !seen.include?(parent)
          seen << parent
          new_nodes << parent
        end
      end
      nodes = new_nodes
    end
    # iterate each node's siblings
    parents.each{it.children.each{}}
  end
  array_of_nodes # not sorted
end
```

### Result
```
Comparison:
                                              child
                                master:      1288.1 i/s 
                   master_sort_improve:      1190.4 i/s - 1.08x  slower
xpath_step_optimize_sort_on_demand_sort_improve:       875.3 i/s - 1.47x  slower
      xpath_step_optimize_sort_improve:       861.3 i/s - 1.50x  slower
    xpath_step_optimize_sort_on_demand:        92.3 i/s - 13.96x  slower
                   xpath_step_optimize:        91.7 i/s - 14.05x  slower
                        sort_on_demand:        90.6 i/s - 14.21x  slower

                                         descendant
                   master_sort_improve:        75.5 i/s 
xpath_step_optimize_sort_on_demand_sort_improve:        75.1 i/s - 1.01x  slower
      xpath_step_optimize_sort_improve:        68.8 i/s - 1.10x  slower
                        sort_on_demand:        21.4 i/s - 3.52x  slower
    xpath_step_optimize_sort_on_demand:        21.4 i/s - 3.52x  slower
                                master:        20.9 i/s - 3.61x  slower
                   xpath_step_optimize:        11.7 i/s - 6.45x  slower

                              descendant-descendant
xpath_step_optimize_sort_on_demand_sort_improve:        47.5 i/s 
      xpath_step_optimize_sort_improve:        41.9 i/s - 1.13x  slower
    xpath_step_optimize_sort_on_demand:        17.9 i/s - 2.65x  slower
                   master_sort_improve:         8.6 i/s - 5.54x  slower
                        sort_on_demand:         6.7 i/s - 7.07x  slower
                   xpath_step_optimize:         6.1 i/s - 7.84x  slower
                                master:         4.6 i/s - 10.24x  slower

                     descendant-descendant-wildcard
xpath_step_optimize_sort_on_demand_sort_improve:       339.5 i/s 
      xpath_step_optimize_sort_improve:       155.9 i/s - 2.18x  slower
    xpath_step_optimize_sort_on_demand:        26.4 i/s - 12.86x  slower
                   master_sort_improve:        10.0 i/s - 33.96x  slower
                        sort_on_demand:         7.7 i/s - 44.30x  slower
                   xpath_step_optimize:         6.8 i/s - 50.06x  slower
                                master:         4.9 i/s - 68.58x  slower

                                ancestor-descendant
xpath_step_optimize_sort_on_demand_sort_improve:       377.9 i/s 
      xpath_step_optimize_sort_improve:       203.7 i/s - 1.85x  slower
    xpath_step_optimize_sort_on_demand:        26.3 i/s - 14.39x  slower
                   xpath_step_optimize:         8.7 i/s - 43.24x  slower
                   master_sort_improve:         7.8 i/s - 48.55x  slower
                        sort_on_demand:         6.3 i/s - 59.93x  slower
                                master:         5.0 i/s - 75.46x  slower

                        preceding-following-sibling
xpath_step_optimize_sort_on_demand_sort_improve:       684.1 i/s 
      xpath_step_optimize_sort_improve:       424.5 i/s - 1.61x  slower
    xpath_step_optimize_sort_on_demand:        85.8 i/s - 7.98x  slower
                   xpath_step_optimize:        23.7 i/s - 28.91x  slower
                   master_sort_improve:        20.9 i/s - 32.72x  slower
                        sort_on_demand:        19.3 i/s - 35.39x  slower
                                master:        13.9 i/s - 49.11x  slower

             preceding-following-sibling-positional
xpath_step_optimize_sort_on_demand_sort_improve:       425.4 i/s 
      xpath_step_optimize_sort_improve:       315.0 i/s - 1.35x  slower
    xpath_step_optimize_sort_on_demand:        84.3 i/s - 5.05x  slower
                   xpath_step_optimize:        23.3 i/s - 18.22x  slower
                   master_sort_improve:         2.1 i/s - 201.38x  slower
                        sort_on_demand:         2.1 i/s - 204.75x  slower
                                master:         1.9 i/s - 222.08x  slower
```

In scenario "child" and "descendant", this PR is slower than master because it adds one additional `sort` call. The difference will be small when `sort` is improved.
In most case, this PR itself does not unleash its full potential because sort is the next bottleneck. Combining with `sort` improvement is important.
The difference of "descendant-descendant" and "descendant-descendant-wildcard" shows that after optimizing sort, the bottleneck will be namespace lookup in qname check for deeply nested xml.

### DIFF
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -338,15 +338,11 @@ module REXML
     end
 
     # Determines if a predicate expression is dependent on the position of nodes.
-    # nil if the predicate is position-independent,
-    # :simple if the predicate is a simple position query that can be optimized in axis scanning
-    # :complex if the predicate is a complex query that might be dependent on the position of nodes
-    def position_dependency(predicate_expr)
-      # [number], [position()=number], [position() < number], [position() > number]
-      return :simple if position_operation(predicate_expr)
-
-      # expressions that contain position-dependent functions
-      return :complex if calls_position_dependent_function?(predicate_expr)
+    # Returns false if the expression is guaranteed to be position-independent.
+    # Returns true if the expression might be position-dependent.
+    def position_dependent?(predicate_expr)
+      # expressions that contain position-dependent functions are position-dependent.
+      return true if calls_position_dependent_function?(predicate_expr)
 
       # Even if expression is followed by path steps, the analysis of
       # position dependency is the same as the expression itself.
@@ -354,32 +350,31 @@ module REXML
       when :union, :or, :and, :eq, :neq, :lt, :lteq, :gt, :gteq, :not
         # Expressions that don't evaluate to a number are position independent
         # if it doesn't contain position-dependent functions.
-        nil
+        false
       when :div, :mod, :mult, :plus, :minus, :neg
-        # expressions that return number. eg. `[position() = @attr + 1]`
-        :complex
+        # expressions that return number. eg. `[@attr + 1]`
+        true
       when :literal
-        # Integer literal is handled in `position_operation(predicate_expr)`.
-        # We treat this as complex(no-optimization) because this is not a normal case
-        # eg. `/foo["hi"]` `/bar[true]` `/baz[3.14]`
-        :complex
+        # Numeric literal is position dependent. String and boolean literal is useless
+        # and not worth optimizing
+        true
       when :variable
         # A variable could resolve to a number at runtime.
         # It's possible to optimize this by checking the actual value of the variable.
-        :complex
+        true
       when :function
         # functions that return number is position dependent. eg. `[position() = string-length(@attr)]`
-        %w[number ceiling round floor string-length sum count].include?(predicate_expr[1]) ? :complex : nil
+        %w[number ceiling round floor string-length sum count].include?(predicate_expr[1])
       when :group
-        position_dependency(predicate_expr[1])
+        position_dependent?(predicate_expr[1])
       when :descendant, :descendant_or_self, :ancestor, :ancestor_or_self,
            :following, :following_sibling, :preceding, :preceding_sibling,
            :document, :child, :self, :parent, :attribute, :namespace
         # paths are position independent. `foo[path[1]]` doesn't depend on the position of `foo`
-        nil
+        false
       else
-        # Every other unhandled expressions are treated as complex for safety
-        :complex
+        # Every other unhandled expressions are treated position dependent for safety
+        true
       end
     end
 
@@ -563,12 +558,12 @@ module REXML
     # If there are multiple position-based predicates or complex position-based predicates,
     # return [position_independent_predicates, nil, nil, complex_predicates]
     def split_positional_predicates(predicates)
-      pre_independent = predicates.take_while {|predicate| position_dependency(predicate).nil? }
+      pre_independent = predicates.take_while {|predicate| !position_dependent?(predicate) }
       predicates = predicates.drop(pre_independent.size)
       return [pre_independent, nil, [], nil] if predicates.empty?
 
       op = position_operation(predicates.first)
-      if op && predicates[1..-1].all? {|predicate| position_dependency(predicate).nil? }
+      if op && predicates[1..-1].all? {|predicate| !position_dependent?(predicate) }
         [pre_independent, op, predicates[1..-1], nil]
       else
         [pre_independent, nil, nil, predicates]
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -199,118 +199,48 @@ module REXML
           nodeset = [nodeset.first.root_node]
         when :self
           nodeset = step(path_stack) do
-            [nodeset]
+            [:iterate_nodesets, [nodeset]]
           end
         when :child
           nodeset = step(path_stack) do
-            child(nodeset)
+            [:iterate_nodesets, child(nodeset)]
           end
         when :literal
           trace(:literal, path_stack, nodeset) if @debug
           return path_stack.shift
         when :attribute
           nodeset = step(path_stack, any_type: :attribute) do
-            nodeset.map do |node|
+            nodesets = nodeset.map do |node|
               next unless node.node_type == :element
               attributes = node.attributes
               next if attributes.empty?
               attributes.each_attribute.to_a
             end.compact
+            [:iterate_nodesets, nodesets]
           end
         when :namespace
           warn 'Namespace axis is not supported in REXML::XPathParser', uplevel: 1
           # TODO: We need to create NamespaceNode class to support this feature
-          nodeset = step(path_stack) { [] }
+          nodeset = step(path_stack) { [:iterate_nodesets, []] }
         when :parent
           nodeset = step(path_stack) do
-            nodesets = []
+            parents = Set.new.compare_by_identity
             nodeset.each do |node|
               if node.node_type == :attribute
                 parent = node.element
               else
                 parent = node.parent
               end
-              nodesets << [parent] if parent
+              parents << parent if parent
             end
-            nodesets
+            [:iterate_nodesets, parents.map {|parent| [parent] }]
           end
-        when :ancestor
-          nodeset = step(path_stack, axis_order: :reverse) do
-            nodesets = []
-            # new_nodes = {}
-            nodeset.each do |node|
-              new_nodeset = []
-              while node.parent
-                node = node.parent
-                # next if new_nodes.key?(node)
-                new_nodeset << node
-                # new_nodes[node] = true
-              end
-              nodesets << new_nodeset unless new_nodeset.empty?
-            end
-            nodesets
-          end
-        when :ancestor_or_self
-          nodeset = step(path_stack, axis_order: :reverse) do
-            nodesets = []
-            # new_nodes = {}
-            nodeset.each do |node|
-              next unless node.node_type == :element
-              new_nodeset = [node]
-              # new_nodes[node] = true
-              while node.parent
-                node = node.parent
-                # next if new_nodes.key?(node)
-                new_nodeset << node
-                # new_nodes[node] = true
-              end
-              nodesets << new_nodeset unless new_nodeset.empty?
-            end
-            nodesets
-          end
-        when :descendant_or_self
+        when :ancestor, :ancestor_or_self,
+            :descendant, :descendant_or_self,
+            :preceding, :preceding_sibling,
+            :following, :following_sibling
           nodeset = step(path_stack) do
-            descendant(nodeset, true)
-          end
-        when :descendant
-          nodeset = step(path_stack) do
-            descendant(nodeset, false)
-          end
-        when :following_sibling
-          nodeset = step(path_stack) do
-            nodeset.map do |node|
-              next unless node.respond_to?(:parent)
-              next if node.parent.nil?
-              all_siblings = node.parent.children
-              current_index = all_siblings.index(node)
-              following_siblings = all_siblings[(current_index + 1)..-1]
-              next if following_siblings.empty?
-              following_siblings
-            end.compact
-          end
-        when :preceding_sibling
-          nodeset = step(path_stack, axis_order: :reverse) do
-            nodeset.map do |node|
-              next unless node.respond_to?(:parent)
-              next if node.parent.nil?
-              all_siblings = node.parent.children
-              current_index = all_siblings.index(node)
-              preceding_siblings = all_siblings[0, current_index].reverse
-              next if preceding_siblings.empty?
-              preceding_siblings
-            end.compact
-          end
-        when :preceding
-          nodeset = step(path_stack, axis_order: :reverse) do
-            nodeset.map do |node|
-              preceding(node)
-            end
-          end
-        when :following
-          nodeset = step(path_stack) do
-            nodeset.map do |node|
-              following(node)
-            end
+            [op, nodeset]
           end
         when :variable
           var_name = path_stack.shift
@@ -382,8 +312,7 @@ module REXML
             target_context[:position] = 1
           end
           args = arguments.dclone.collect do |arg|
-            result = expr(arg, nodeset, target_context)
-            result
+            expr(arg, nodeset, target_context)
           end
           @functions.context = target_context
           return @functions.send(func_name, *args)
@@ -394,7 +323,7 @@ module REXML
             # If result is a nodeset, apply following predicates
             path_stack.unshift(:node)
             nodeset = step(path_stack) do
-              [result]
+              [:iterate_nodesets, [result]]
             end
           else
             return result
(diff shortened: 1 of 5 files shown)


lib/rexml/xpath_parser.rb
```diff
@@ -449,138 +363,358 @@ def expr( path_stack, nodeset, context=nil )
...
+        end
+        ordered = sort(seen.keys, order)
+        ordered.map.with_index(1) do |raw_node, position|
+          XPathNode.new(raw_node, position: position)
         end
-        new_nodeset
       ensure
-        leave(:step, path_stack, new_nodeset) if @debug
+        leave(:step, path_stack, ordered) if @debug
       end
```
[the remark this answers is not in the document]
tompng: Referencing unassigned local lexically after local variable definition result in nil, not NameError.

lib/rexml/xpath_parser.rb
```diff
@@ -245,113 +241,31 @@ def expr( path_stack, nodeset, context=nil )
                 end
               end
             end
-            nodesets
+            # Not working at all, so just return an empty nodesets for now.
+            # Needs Namespace-node class
+            [:iterate_raw_nodesets, []]
```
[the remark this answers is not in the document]
tompng: Previous behavior was raising NoMethodError

lib/rexml/xpath_parser.rb
```diff
@@ -449,138 +373,358 @@ def expr( path_stack, nodeset, context=nil )
...
+                matched << node if index >= value
+              end
+              index += 1
+            end
+            if anchors.key?(node)
+              anchor_indexes[index] = true
+              last_anchor = index
+            end
+          end
+          matched
```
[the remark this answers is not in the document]
tompng: No. In `>` case, index from the first anchor(always zero) is the possible maximum index among relative index from several anchor nodes. I modified the code a bit, and added source code comment that describe this.

lib/rexml/xpath_parser.rb
```diff
@@ -449,138 +373,358 @@ def expr( path_stack, nodeset, context=nil )
...
+        nodesets.each do |nodeset|
+          nodeset.each do |node|
+            raw_node = node.respond_to?(:raw_node) ? node.raw_node : node
+            seen[raw_node] = true
+          end
+        end
+        ordered = sort(seen.keys)
+        ordered.map.with_index(1) do |raw_node, position|
+          XPathNode.new(raw_node, position: position)
         end
```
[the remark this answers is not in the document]
tompng: REXML previously skipped nodesets ordering when `nodesets.size == 1`, but it turned out that it was a bug. `nodesets.first` is sometimes sorted in reverse-document-order while the order should be document-order.

We can reduce sort in the future such as:
- Add an ordered/reverse-ordered/unordered flag to nodeset
- Sort only on the final result and before applying filter predicates

But this kind of separate optimization should be out of scope of this PR.

lib/rexml/xpath_parser.rb
```diff
@@ -460,178 +357,377 @@ def expr( path_stack, nodeset, context=nil )
...
+          children = children.reverse if reverse
+          children.drop_while {|child| !sets.key?(child) }.drop(1)
+        end.select(&tester)
+      when :nodesets
+        nodesets = nodeset.map do |node|
+          parent = node.parent
+          index = parent.children.index(node)
+          reverse ? parent.children[0...index].reverse : parent.children[index + 1..-1]
+        end
+        non_optimized_nodesets_select(nodesets, tester, selector)
```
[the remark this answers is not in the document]
tompng: No. This is a non-optimizable slow path that needs O(n^2) calculation cost just like before.

naitoh:
@tompng

> Remove XPathNode

- Since this relates to https://github.com/ruby/rexml/pull/325, please update the description.

tompng:
Done removing that section

lib/rexml/xpath_parser.rb
```diff
@@ -408,120 +337,352 @@ def expr( path_stack, nodeset, context=nil )
...
+        operator, value = selector
+        nodes = nodesets.flatten
+        nodes =
+          case operator
+          when :==
+            nodesets.map {|nodeset| nodeset[value - 1] if value >= 1 }.compact
+          when :<
+            nodesets.map {|nodeset| nodeset[0...value - 1] if value >= 1 }.compact.flatten
+          when :>
+            nodesets.map {|nodeset| value < 0 ? nodeset : nodeset[value..-1] }.flatten
```
naitoh:
```suggestion
            nodesets.flat_map {|nodeset| value < 0 ? nodeset : nodeset.drop(value) }
```

If `position() > N` is out of bounds, a `nil` value is introduced, causing a crash.

* before
```ruby
> xml = <<-XML
        <root>
          <a/>
          <b/>
          <c/>
          <d/>
        </root>
      XML
> REXML::XPath.match(REXML::Document.new(xml), "/root/*[position() > 10]")
=> []
```
* after(this PR)
```ruby
=> [nil]
```
* before
```ruby
> xml2 = <<-XML
<a>
  <b>
    <c/>
  </b>
  <b>
    <c/>
    <c/>
    <c/>
  </b>
</a>
      XML
> REXML::XPath.match(REXML::Document.new(xml2), "/a/b/c[position() > 2]")
=> [<c/>]
```
* after(this PR)
```ruby
rexml/lib/rexml/xpath_parser.rb:750:in 'block in REXML::XPathParser#sort': undefined method 'node_type' for nil (NoMethodError)

        np = node.node_type == :attribute ? node.element : node
                 ^^^^^^^^^^
	from rexml/lib/rexml/xpath_parser.rb:748:in 'Array#each'
	from rexml/lib/rexml/xpath_parser.rb:748:in 'REXML::XPathParser#sort'
	from rexml/lib/rexml/xpath_parser.rb:604:in 'REXML::XPathParser#step'
	from rexml/lib/rexml/xpath_parser.rb:205:in 'REXML::XPathParser#expr'
	from rexml/lib/rexml/xpath_parser.rb:156:in 'REXML::XPathParser#match'
	from rexml/lib/rexml/xpath_parser.rb:91:in 'block in REXML::XPathParser#parse'
	from rexml/lib/rexml/document.rb:459:in 'REXML::Document#enable_cache'
	from rexml/lib/rexml/xpath_parser.rb:90:in 'REXML::XPathParser#parse'
	from rexml/lib/rexml/xpath.rb:66:in 'REXML::XPath.match'
	from (irb):17:in '<main>'
```

lib/rexml/xpath_parser.rb
```diff
@@ -408,120 +337,352 @@ def expr( path_stack, nodeset, context=nil )
...
+      case selector
+      when :nodesets
+        nodesets
+      when :uniq
+        seen = Set.new.compare_by_identity
+        nodesets.flatten.each {|node| seen << node }
+        seen.to_a
+      else
+        operator, value = selector
+        nodes = nodesets.flatten
```
naitoh: `nodes` is not being used.

lib/rexml/xpath_parser.rb
```diff
@@ -408,120 +337,352 @@ def expr( path_stack, nodeset, context=nil )
       leave(:expr, path_stack, nodeset) if @debug
     end
 
-    def step(path_stack, any_type: :element, axis_order: :forward)
-      nodesets = yield
+    # Determines if a predicate expression is dependent on the position of nodes.
+    # nil if the predicate is position-independent,
+    # :simple if the predicate is a simple position query that can be optimized in axis scanning
+    # :complex if the predicate is a complex query that might be dependent on the position of nodes
+    def position_dependency(predicate_expr)
```
naitoh: There is no handling for cases where `predicate_expr[0] == :group`.

* `/root/*[2]`
```ruby
> predicate_expr
[:literal, 2]
```
* `/root/*[(2)]`
```ruby
> predicate_expr
[:group, [:literal, 2]]
```

* before
```ruby
> xml = <<-XML
        <root>
          <a/>
          <b/>
          <c/>
          <d/>
        </root>
      XML
>  REXML::XPath.match(REXML::Document.new(xml), "/root/*[(2)]")
=> [<b/>]
```
* after(this PR)
```ruby
=> []
```

lib/rexml/xpath_parser.rb
tompng: Fixed and add test for it

lib/rexml/xpath_parser.rb
tompng: Changed it to a case when that fallbacks to `:complex` so that we don't handle complex case as position-independent

lib/rexml/xpath_parser.rb
naitoh, line 771: The order of the nodes directly under the `Document` is incorrect.

```suggestion
        while np.parent and (np.parent.node_type == :element or np.parent.node_type == :document)
```

* before
```ruby
> xml = '<!--c1--><!--c2--><!--c3--><root><x/><y/></root><!--c4--><!--c5-->'
> REXML::XPath.match(REXML::Document.new(xml), "/root/preceding-sibling::comment()").map(&:string)
=> ["c1", "c2", "c3"]
> REXML::XPath.match(REXML::Document.new(xml), "/root/preceding::node()").map(&:string)
=> ["c1", "c2", "c3"]
> REXML::XPath.match(REXML::Document.new(xml), "/root/following-sibling::comment()").map(&:string)
=> ["c4", "c5"]
> REXML::XPath.match(REXML::Document.new(xml), "/descendant-or-self::node()")
=> 
[<UNDEFINED> ... </>,
 <?xml ... ?>,
 #<REXML::Comment:0x0000000124a9b530 @parent=<UNDEFINED> ... </>, @string="c1">,
 #<REXML::Comment:0x0000000124a9b198 @parent=<UNDEFINED> ... </>, @string="c2">,
 #<REXML::Comment:0x0000000124a9af90 @parent=<UNDEFINED> ... </>, @string="c3">,
 <root> ... </>,
 <x/>,
 <y/>,
 #<REXML::Comment:0x0000000124a9a180 @parent=<UNDEFINED> ... </>, @string="c4">,
 #<REXML::Comment:0x0000000124a99fc8 @parent=<UNDEFINED> ... </>, @string="c5">]
```

* after(this PR)
```ruby
> xml = '<!--c1--><!--c2--><!--c3--><root><x/><y/></root><!--c4--><!--c5-->'
> REXML::XPath.match(REXML::Document.new(xml), "/root/preceding-sibling::comment()").map(&:string)
=> ["c3", "c2", "c1"]
> REXML::XPath.match(REXML::Document.new(xml), "/root/preceding::node()").map(&:string)
=> ["c3", "c2", "c1"]
> REXML::XPath.match(REXML::Document.new(xml), "/root/following-sibling::comment()").map(&:string)
=> ["c4", "c5"]
> REXML::XPath.match(REXML::Document.new(xml), "/descendant-or-self::node()")
=> 
[#<REXML::Comment:0x0000000124546aa8 @parent=<UNDEFINED> ... </>, @string="c5">,
 #<REXML::Comment:0x0000000124547868 @parent=<UNDEFINED> ... </>, @string="c1">,
 #<REXML::Comment:0x0000000124547660 @parent=<UNDEFINED> ... </>, @string="c2">,
 #<REXML::Comment:0x0000000124547458 @parent=<UNDEFINED> ... </>, @string="c3">,
 <root> ... </>,
 #<REXML::Comment:0x0000000124546c60 @parent=<UNDEFINED> ... </>, @string="c4">,
 <UNDEFINED> ... </>,
 <x/>,
 <y/>]
```

tompng: It's an existing bug of `sort`

```ruby
# master(7d9e7c2e08d81b688ab6a9d0d1b329b619d1ca55)
xml = '<!--c1--><!--c2--><!--c3--><foo><x/><y/><inner><!--comment--><foo/></inner></foo><!--c4--><!--c5-->'

REXML::XPath.match(doc, "/foo/preceding-sibling::comment()").map(&:string)
# => ["c1", "c2", "c3"] # This was OK because sort is skipped when nodesets.size==1 in master
REXML::XPath.match(doc, "//foo/preceding-sibling::comment()").map(&:string)
# => ["c3", "c2", "c1", "comment"] # If there's more than two `//foo` in the document, it'll run into the bug.
```

There's at least three bugs related to `sort`
1. Unable to sort root comments because all sort keys are `[]`
2. Sort key of attribute is the same as attribute.element's key
3. Sorting union of attributes and elements won't work because of 2, but before that, the sorting of union wasn't even performed

```ruby
# master and in v3.4.4

# Bug 1
xml = '<foo><foo/><!--c0--></foo>' + 100.times.map{"<!--c#{it+1}-->"}.join
REXML::XPath.match(doc, "//foo/following-sibling::comment()").map(&:string)
# => ["c77", "c2", "c3", "c4", ..., "c75", "c76", "c1", "c78", "c79", ..., "c99", "c100", "c0"]

# Bug 2
xml = "<root>#{2.times.map{|i| "<foo #{50.times.map{"a#{50*i+it}='true'"}.join(' ')}/>" }.join}</root>"
REXML::XPath.match(doc, "//foo/attribute::*").map(&:name)
# => ["a0", "a2", "a3", ..., "a48", "a49", "a1", "a99", "a50", "a51", "a52", ..., "a97", "a98"]

# Bug 3
xml = "<root><a attr1='1' attr2='2'/><b/><c/></root>"
REXML::XPath.match(doc, "//c | //attribute::attr2 | //a | //attribute::attr1 | //b").map(&:name)
# => ["c", "attr2", "a", "attr1", "b"] # not evne sorted
# => ["attr2", "a", "attr1", "b", "c"] # Wrong even after changing to `sort(left | right)`
```

I think it can be fix it in a separate pull request, or perhaps leave it until rewriting sort with a performant algorithm.

naitoh: OK, I see.

naitoh (approved):
I understand this PR scan strategy.
Thank you!

PR #317, opened 2026-05-22
Partially support parenthesized xpath

### LINKED ISSUE #25: Node set predicate not working

tompng:
Fix #25, and also reverts #122.

Adds a new path_stack operator `:group`

### DIFF
--- a/lib/rexml/parsers/xpathparser.rb
+++ b/lib/rexml/parsers/xpathparser.rb
@@ -654,7 +654,7 @@ module REXML
           contents = contents[1..-2]
           n = []
           OrExpr( contents, n )
-          parsed.concat(n)
+          parsed.push(:group, n)
         end
         path
       end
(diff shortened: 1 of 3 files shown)


naitoh:
Thanks!

naitoh (approved):

Issue #319, opened 2026-05-26
Inconsistent XPath node ordering and non-compliance with XPath 1.0


tompng:
## Inconsistent ordering bug
Ordering of the final result is inconsistent. This is just a bug. We need to standardize on one or the other.

```ruby
doc = REXML::Document.new('<a><b><c><d id="1"><d id="2"/></d></c></b></a>')

REXML::XPath.first(doc, '//d/ancestor::*').name
#=> "a" (document order)
REXML::XPath.first(doc, '//d[@id="2"]/ancestor::*').name
#=> "d" (reverse document order)

REXML::XPath.match(doc, '//d/ancestor::*').map(&:name)
#=> ["a", "b", "c", "d"] (document order)
REXML::XPath.match(doc, '//d[@id="2"]/ancestor::*').map(&:name)
#=> ["d", "c", "b", "a"] (reverse document order)

REXML::XPath.first(doc, '//d/ancestor-or-self::*').name
#=> "a" (document order)
REXML::XPath.first(doc, '//d[@id="2"]/ancestor-or-self::*').name
#=> "d" (reverse-document order)
REXML::XPath.match(doc, '//d/ancestor-or-self::*').map(&:name)
#=> ["a", "b", "c", "d", "d"] (document order)
REXML::XPath.match(doc, '//d[@id="2"]/ancestor-or-self::*').map(&:name)
#=> ["d", "d", "c", "b", "a"] (reverse document order)
```

## XPath 1.0 requires document order

https://www.w3.org/TR/1999/REC-xpath-19991116/#node-sets

Nodeset are basically unordered

> node-set (an unordered collection of nodes without duplicates)

Axis has order

> An axis is either a forward axis or a reverse axis
> The proximity position of a member of a node-set with respect to an axis is defined to be the position of the node in the node-set ordered in document order if the axis is a forward axis and ordered in reverse document order if the axis is a reverse axis.

Positional predicates applied to axis uses axis order

> The meaning of a Predicate depends crucially on which axis applies.
> For example, preceding::foo[1] returns the first foo element in reverse document order, because the axis that applies to the [1] predicate is the preceding axis;

After applying predicates to reverse axis, it is unordered, and may need to be sorted in document order before applying another predicates

> by contrast, (preceding::foo)[1] returns the first foo element in document order, because the axis that applies to the [1] predicate is the child axis.
[...]

PR #320, opened 2026-05-26
XPath nodeset always in document-order

tompng:
Fixes problems described in #319
Change nodeset ordering to match XPath 1.0 spec
Changes `XPath#first` `XPath#match` ordering

In XPath 1.0 spec, nodeset are unordered set, but many operations are required to use document ordered.
Functions such as `local-name` should use first node in document order.
Filter expressions such as `(preceding::foo)[1]` need to use document-order.
JavaScript's `XPathResult.ORDERED_NODE_ITERATOR_TYPE` `XPathResult.FIRST_ORDERED_NODE_TYPE` always orders in document order.
This commit will make REXML match to this ordering, and also fixes inconsistent ordering bug: same xpath may return document-ordered nodes or reverse-document-ordered nodes depending on `nodesets.size`.

### DIFF
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -262,7 +262,7 @@ module REXML
             nodesets
           end
         when :ancestor
-          nodeset = step(path_stack) do
+          nodeset = step(path_stack, order: :reverse) do
             nodesets = []
             # new_nodes = {}
             nodeset.each do |node|
@@ -280,7 +280,7 @@ module REXML
             nodesets
           end
         when :ancestor_or_self
-          nodeset = step(path_stack) do
+          nodeset = step(path_stack, order: :reverse) do
             nodesets = []
             # new_nodes = {}
             nodeset.each do |node|
@@ -471,7 +471,7 @@ module REXML
           nodesets = evaluate_predicate(predicate_expression, nodesets)
         end
         if nodesets.size == 1
-          ordered_nodeset = nodesets[0]
+          ordered_nodeset = order == :forward ? nodesets.first : nodesets.first.reverse
         else
           seen = {}.compare_by_identity
           raw_nodes = []
@@ -483,7 +483,7 @@ module REXML
               raw_nodes << raw_node
             end
           end
-          ordered_nodeset = sort(raw_nodes, order)
+          ordered_nodeset = sort(raw_nodes)
         end
         new_nodeset = []
         ordered_nodeset.each do |node|
@@ -667,7 +667,7 @@ module REXML
     # in and out of function calls.  If I knew what the index of the nodes was,
     # I wouldn't have to do this.  Maybe add a document IDX for each node?
     # Problems with mutable documents.  Or, rewrite everything.
-    def sort(array_of_nodes, order)
+    def sort(array_of_nodes)
       new_arry = []
       array_of_nodes.each { |node|
         node_idx = []
@@ -679,11 +679,7 @@ module REXML
         new_arry << [ node_idx.reverse, node ]
       }
       ordered = new_arry.sort_by do |index, node|
-        if order == :forward
-          index
-        else
-          index.map(&:-@)
-        end
+        index
       end
       ordered.collect do |_index, node|
         node
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -262,7 +262,7 @@ module REXML
             nodesets
           end
         when :ancestor
-          nodeset = step(path_stack, order: :reverse) do
+          nodeset = step(path_stack, axis_order: :reverse) do
             nodesets = []
             # new_nodes = {}
             nodeset.each do |node|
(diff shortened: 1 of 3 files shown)


lib/rexml/xpath_parser.rb
```diff
@@ -459,21 +459,19 @@ def step(path_stack, any_type: :element, order: :forward)
           predicate_expression = path_stack.shift.dclone
           nodesets = evaluate_predicate(predicate_expression, nodesets)
         end
-        if nodesets.size == 1
-          ordered_nodeset = nodesets[0]
-        else
```
naitoh, line 475: @tompng 
Thank you for this PR.

But, performance has decreased because the optimization for `nodesets.size == 1` has been removed.
Is there any way to improve this?


```bash
$ git checkout master
$ rake build
rexml 3.4.5 built to pkg/rexml-3.4.5.gem.
$ gem install pkg/rexml-3.4.5.gem
$ git checkout no_reverse_order
$ benchmark-driver benchmark/xpath_no_reverse_order.yaml
                     master(a98066c)       after  master(a98066c,YJIT)  after(YJIT) 
               child         280.227      19.016               544.496       31.084 i/s -     100.000 times in 0.356853s 5.258603s 0.183656s 3.217105s
          descendant          18.445       9.682                28.523       15.258 i/s -     100.000 times in 5.421532s 10.327928s 3.505883s 6.554030s

Comparison:
                            child
master(a98066c,YJIT):       544.5 i/s 
     master(a98066c):       280.2 i/s - 1.94x  slower
         after(YJIT):        31.1 i/s - 17.52x  slower
               after:        19.0 i/s - 28.63x  slower

                       descendant
master(a98066c,YJIT):        28.5 i/s 
     master(a98066c):        18.4 i/s - 1.55x  slower
         after(YJIT):        15.3 i/s - 1.87x  slower
               after:         9.7 i/s - 2.95x  slower
```

- benchmark/xpath_no_reverse_order.yaml
```yaml
loop_count: 100
contexts:
  - name: master(a98066c)
    gems:
      rexml: 3.4.5
    require: false
    prelude: require 'rexml'
  - name: after
    prelude: |
      require 'rexml'
  - name: master(a98066c,YJIT)
    gems:
      rexml: 3.4.5
    require: false
    prelude: |
      require 'rexml'
      RubyVM::YJIT.enable
  - name: after(YJIT)
    prelude: |
      require 'rexml'
      RubyVM::YJIT.enable

prelude: |
  require "rexml"
  xml = "<root>" + (1..2000).map { |i| "<item id='#{i}'/>" }.join + "</root>"

benchmark:
  child: REXML::XPath.match(doc, "/root/item")
  descendant: REXML::XPath.match(doc, "//item")
```

tompng: > Is there any way to improve this?

Yes, there are several ways.

- A. Improve performance of sort
  - Should be done in a separate PR

- B. Mark nodesets ordering
  - `ordered_nodeset = marked_as_ordered? ? nodesets.first : nodesets.first.reverse`
  - Ordering is currently `:ordered | :reversed` but in #315 we need to add `:unordered`
  - Conflicts with C

- C. Delay nodeset ordering on demand
  - Sort the final result and in some functions such `number(xpath)`
  - Requires predicate position-independency analytics, requires sorted nodeset or not
  - After doing this, optimization of B needs to be extended. We need to re-implement B in a different way again.

I implemented B with changing the meaning of `order:` keyword parameter

- Before: key to specify the output order: `:forward | :reverse`
- After: key to specify the state of nodesets ordering: `:ordered | :reversed | :unordered(future)`

Perhaps this approach will be a blocker implementing C(delayed nodeset ordering), and optimizing `XPathParser#step` (#315) but we can consider it later.

naitoh: > I implemented B with changing the meaning of `order:` keyword parameter
> 
>     * Before: key to specify the output order: `:forward | :reverse`
> 
>     * After: key to specify the state of nodesets ordering: `:ordered | :reversed | :unordered(future)`
> 
> 
> Perhaps this approach will be a blocker implementing C(delayed nodeset ordering), and optimizing `XPathParser#step` (#315) but we can consider it later.

@tompng 

Thank you.
I also think B is the best option, but can we keep `:forward | :reverse` in this PR?

I don't want to include `:ordered | :reversed | :unordered(future)` in this PR, because I can't review them at this time, as they are scheduled to be changed later.

I would like to include changes to `:ordered | :reversed | :unordered(future)` in the pull request when they are actually implemented.

tompng: Updated. Copilot suggested to rename the keyword parameter to `:axis_order`, what do you think?

naitoh: Thank you for your update.

Also, `:axis_order` is easier to understand.
Please change it to `:axis_order`.

tompng: Done!

naitoh (approved):
Thanks!

PR #323, opened 2026-06-05
Remove XPath namespace axis handling part which is not working at all

tompng:
Namespace axis should return a nodeset of Namespace Node, but Namespace Node class has not even been defined yet.
So namespace axis is completely unimplemented.
Right now, this incomplete code is just getting in the way of refactoring.

```ruby
REXML::XPath.match(REXML::Document.new('<root/>'), '//*/namespace::*')
# before:
# => in REXML::XPathParser#node_test': undefined method 'raw_node' for an instance of Array (NoMethodError)

# after:
# warning: Namespace axis is not supported in REXML::XPathParser
# => []
```

Other options:
- Silently return empty nodeset
- Always raise error with a message

### DIFF
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -223,30 +223,9 @@ module REXML
             nodesets
           end
         when :namespace
-          pre_defined_namespaces = {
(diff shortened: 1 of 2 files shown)


naitoh (approved):
Thanks!

PR #324, opened 2026-06-05
Fix XPath normalize-space function

tompng:
It should return normalized string of the first node, just like other functions such as `string()` and `number()`

```ruby
doc = Nokogiri::XML.parse(<<~XML)
  <a><b>breakfast     boosts\t\t
  
  concentration   </b><c>
  Coffee beans
  aroma
  </c><d>        Dessert
  \t\t    after    dinner</d></a>
XML
doc.xpath("normalize-space(//text())")
#=> "breakfast boosts concentration"
```

This bug is a blocker of implementing delayed nodeset ordering.
```ruby
def match(path_stack, node)
  nodeset = [XPathNode.new(node, position: 1)]
  result = expr(path_stack, nodeset)
  case result
  when Array # nodeset ← HERE, normalize-space function returned array of string
    unnode(result).uniq # Changing to sort(result) failed
  else
    [result]
  end
end
```

### DIFF
--- a/lib/rexml/functions.rb
+++ b/lib/rexml/functions.rb
@@ -262,13 +262,8 @@ module REXML
       string(string).length
     end
 
-    def Functions::normalize_space( string=nil )
-      string = string(@@context[:node]) if string.nil?
-      if string.kind_of? Array
-        string.collect{|x| x.to_s.strip.gsub(/\s+/um, ' ') if x}
-      else
-        string.to_s.strip.gsub(/\s+/um, ' ')
-      end
+    def Functions::normalize_space( object=@@context[:node] )
+      string(object).strip.gsub(/\s+/um, ' ')
     end
 
     # This is entirely Mike Stok's beast
--- a/test/functions/test_base.rb
+++ b/test/functions/test_base.rb
@@ -242,13 +242,18 @@ Coffee beans
 </c><d>        Dessert
  \t\t    after    dinner</d></a>
       XML
-      normalized_texts = REXML::XPath.each(REXML::Document.new(source), "normalize-space(//text())").to_a
-      assert_equal([
-                     "breakfast boosts concentration",
-                     "Coffee beans aroma",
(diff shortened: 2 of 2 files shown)


test/functions/test_base.rb
```diff
@@ -242,13 +242,12 @@ def test_normalize_space_strings
...
  \t\t    after    dinner</d></a>
       XML
-      normalized_texts = REXML::XPath.each(REXML::Document.new(source), "normalize-space(//text())").to_a
-      assert_equal([
-                     "breakfast boosts concentration",
-                     "Coffee beans aroma",
-                     "Dessert after dinner",
-                   ],
-                   normalized_texts)
+      normalized_text = REXML::XPath.each(REXML::Document.new(source), "normalize-space(//text())").to_a
```
naitoh: normalized_text has not been tested.

lib/rexml/functions.rb
```diff
@@ -263,12 +263,7 @@ def Functions::string_length( string )
     end
 
     def Functions::normalize_space( string=nil )
```
naitoh: If it is called without any arguments, as in `normalize-space()`, this causes a problem.

* master
```
require "nokogiri"
require "rexml"

xml = <<-XML
        <root>
          <item>  apple  </item>
          <item>  orange  </item>
          <item>  banana  </item>
        </root>
      XML
puts Nokogiri::XML.parse(xml).xpath("//item/text()[normalize-space()='orange']")
#=> "  orange  "
puts REXML::XPath.first(doc, "//item/text()[normalize-space()='orange']")
#=> "  orange  "
```
* this PR

```
puts REXML::XPath.first(doc, "//item/text()[normalize-space()='orange']")
#=> ""
```

tompng: Nice catch! Fixed and test added.

naitoh (approved):
Thanks!

PR #325, opened 2026-06-05
Eliminate XPathNode

tompng:
XPathNode requires :position, but in most case, this :position is not used at all, and just doing useless packing/unpacking.

Extracted from #315


naitoh (approved):
I think it’s much easier to read now.
Thank you!

PR #326, opened 2026-06-05
Primitive value should't be wrapped in an array

tompng:
Similar to #324
Values should be nodeset(Array) or primitive, not `[primitive]`

Also removes these workarounds related to primitive wrapped in array
```ruby
# Workaround to handle single primitive value wrapped in an array
result = result[0] if result.kind_of? Array and result.length == 1

# Workaround to handle multiple primitive values wrapped in an array
# Arrays are always nodeset, so `if result.size > 0` is enough
if result.size > 0 and result.inject(false) {|k,s| s or k}
```

It works even wrapping with an array, but leaving it will be a blocker of implementing delayed nodeset ordering.

### DIFF
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -313,7 +313,7 @@ module REXML
           end
         when :variable
           var_name = path_stack.shift
-          return [@variables[var_name]]
+          return @variables[var_name]
 
         when :eq, :neq, :lt, :lteq, :gt, :gteq
           left = expr( path_stack.shift, nodeset.dup, context )
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -533,16 +533,13 @@ module REXML
           subcontext[:position] = position
           result = expr(expression.dclone, [node], subcontext)
           trace(:predicate_evaluate, expression, node, subcontext, result) if @debug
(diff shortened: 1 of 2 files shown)


lib/rexml/xpath_parser.rb
```diff
@@ -313,7 +313,7 @@ def expr( path_stack, nodeset, context=nil )
           end
         when :variable
           var_name = path_stack.shift
-          return [@variables[var_name]]
+          return @variables[var_name]
```
tompng, line 316: Returning `nil` may cause unintentional behavior. However, it was returning `[nil]` before, and it may also cause problem, so basically nothing has changed.

In XPath 1.0, only `string | number | boolean | nodeset` are allowed.
In Nokogiri, using undefined variable will raise `Undefined variable: $foo (Nokogiri::XML::XPath::SyntaxError)`.
In JavaScript's `document.execute`, all variables are bound to empty string.

Ideally, REXML should validate variables somewhere, and need to decide how to handle undefined variables: fallback or raise. Although changing/deciding this may be out of scope of this pull request.

Ref: https://www.w3.org/TR/1999/REC-xpath-19991116/#section-Basics
> A VariableReference evaluates to the value to which the variable name is bound in the set of variable bindings in the context. It is an error if the variable name is not bound to any value in the set of variable bindings in the expression context.

naitoh (approved):
Thanks!

PR #327, opened 2026-06-05
Remove unused xpath function text() and processing-instruction()

tompng:
Related to #326
Functions must return nodeset or a primitive value. Functions::text() returning an array of string is a bug to be fixed. Although, `text()` in xpath is resolved as `[:child, :text]`, it's not a function, so Functions::text is unused.

Also removes `Functions::processing_instruction`. These are node-test, not functions.

### DIFF
--- a/lib/rexml/functions.rb
+++ b/lib/rexml/functions.rb
@@ -419,10 +419,6 @@ module REXML
       end
     end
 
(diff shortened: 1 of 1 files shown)


naitoh (approved):
Thank you!

PR #328, opened 2026-06-06
Implement XPath abbreviate of parentheses and fix some bugs

tompng:
Implement `abbreviate` of parenthesized `:group` added in #317

Also fixes these bugs so that we can test parenthesized xpaths such as `(1+2)*3`
```ruby
parser = REXML::Parsers::XPathParser.new
parser.abbreviate parser.parse("count(/a)") #=> NoMethodError
parser.abbreviate parser.parse("-1") #=> NoMethodError
parser.abbreviate parser.parse("1+2") #=> NoMethodError
parser.abbreviate parser.parse("/a[1+2]") #=> "/a[1 plus 2]" (XPath doesn't have `plus` keyword)
```

### DIFF
--- a/lib/rexml/parsers/xpathparser.rb
+++ b/lib/rexml/parsers/xpathparser.rb
@@ -103,14 +103,16 @@ module REXML
             component << ']'
           when :document
             components << ""
-          when :function
-            component << parsed.shift
-            component << "( "
-            component << predicate_to_path(parsed.shift[0]) {|x| abbreviate(x)}
-            component << " )"
-          when :literal
-            component << quote_literal(parsed.shift)
+          when :function, :literal, :group, :neg, :and, :or, :mult, :plus, :minus, :neq, :eq, :lt, :gt, :lteq, :gteq, :div, :mod, :union
+            component = ''
+            components << component
+            parsed.unshift(op)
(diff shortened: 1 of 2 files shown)


lib/rexml/parsers/xpathparser.rb
```diff
@@ -103,14 +103,16 @@ def abbreviate(path_or_parsed)
...
           when :document
             components << ""
-          when :function
-            component << parsed.shift
-            component << "( "
-            component << predicate_to_path(parsed.shift[0]) {|x| abbreviate(x)}
-            component << " )"
-          when :literal
-            component << quote_literal(parsed.shift)
+          when :function, :literal, :group, :neg, :and, :or, :mult, :plus, :minus, :neq, :eq, :lt, :gt, :lteq, :gteq, :div, :mod, :union
```
tompng, line 106: This is not good, but we need to do this to avoid infinite recursion of unknown operator.
`predicate_to_path` calls `abbreviate` if `predicate_to_path` doesn't know the operator.

Some possible workaround for this may break `XPathParser#expand` which is not tested and the spec is unknown.

naitoh (approved):
Thanks!

PR #333, opened 2026-06-14
Change namespace cache strategy

tompng:
Builds on top of #335

Even if namespace lookup is cached, namespace lookup was slow for deeply nested XML nodes.
Namespaces are cached in `document`, but `document` lookup from a deep node costs `O(n)`.

We can't cache it in a document or root of the node, because we also need to cache document lookup result.
It's not good to cache to node itself, for maintainability reason, cache invalidation, etc.
In this PR, Hash for caching node's namespace/namespaces, and cached namespace attributes defined in attlist decl can be injected through method parameters.

Benchmark
```ruby
irb> doc = REXML::Document.new('<a attr="1">'*1000+'</a>'*1000)
irb> measure
irb> REXML::XPath.match(doc, '//a');
# Master:                             0.2791398s
# Master(XPathParser#sort disabled):  0.1138055s
# This PR:                            0.1743505s
# This PR(XPathParser#sort disabled): 0.0046388s
```

Time consumption was: Scan: 0.0046s (1.8%), Sort: 0.17s (60%), Namespace lookup: 0.11s (38%). This PR removes the 38% cost.

Cache through parameters also speeds up bare `namespace` call on deeply nested element.
```ruby
irb> doc = REXML::Document.new('<a>'*1000+'</a>'*1000)
irb> node = doc; node = node.first while node.first
irb> measure
irb> node.namespace
# Master:  0.079568s O(N^2)
# This PR: 0.003091s O(N)
```

### DIFF
--- a/lib/rexml/doctype.rb
+++ b/lib/rexml/doctype.rb
@@ -123,7 +123,7 @@ module REXML
       _attlist_mappings[element]&.[](attribute)
     end
 
-    def _attlist_mappings
+    def _attlist_mappings # :nodoc:
       mapping = {}
       grep(AttlistDecl).each do |child|
         raw_attributes = mapping[child.element_name] ||= {}
--- a/lib/rexml/doctype.rb
+++ b/lib/rexml/doctype.rb
@@ -113,27 +113,26 @@ module REXML
     end
 
     def attributes_of element
-      attribute_declarations_of(element).map do |key, value|
+      raw_attributes = _attlist_mappings[element] || {}
+      raw_attributes.map do |key, value|
         Attribute.new(key, value)
       end
     end
 
     def attribute_of element, attribute
-      attribute_declarations_of(element)[attribute]
+      _attlist_mappings[element]&.[](attribute)
     end
 
-    def attribute_declarations_of(name)
-      raw_attributes = {}
-      decls = select do |child|
-        child.kind_of?(AttlistDecl) && child.element_name == name
-      end
-      decls.each do |child|
+    def _attlist_mappings
+      mapping = {}
+      grep(AttlistDecl).each do |child|
+        raw_attributes = mapping[child.element_name] ||= {}
         child.each do |key, val|
           # First declaration wins
           raw_attributes[key] = val unless raw_attributes.key? key
         end
       end
-      raw_attributes
+      mapping
     end
 
     def clone
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -2527,7 +2527,8 @@ module REXML
     # the element's attribute takes precedence.
     def each_effective_attribute(attlist_mappings = nil)
       return to_enum(__method__, attlist_mappings) unless block_given?
-      attributes = attlist_attributes(attlist_mappings).merge(self)
+      attlist = attlist_attributes(attlist_mappings)
+      attributes = attlist.empty? ? self : attlist.merge(self)
       attributes.each_value do |attribute|
         yield attribute
       end
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -588,12 +588,7 @@ module REXML
     #   d.elements['//c'].namespaces # => {"x"=>"1", "y"=>"2", "z"=>"3"}
     #
     def namespaces
-      namespaces_cache = document&.__send__(:namespaces_cache)
-      if namespaces_cache
-        namespaces_cache[self] ||= calculate_namespaces
-      else
-        calculate_namespaces
-      end
+      _calculate_namespaces
     end
 
     # :call-seq:
@@ -618,13 +613,10 @@ module REXML
     #
     def namespace(prefix=nil)
       if prefix.nil?
-        prefix = prefix()
+        _namespace_internal
+      else
+        _namespace_lookup_internal(prefix)
       end
-      prefix = (prefix == '') ? 'xmlns' : prefix.delete_prefix("xmlns:")
-      ns = namespaces[prefix]
-
-      ns = '' if ns.nil? and prefix == 'xmlns'
-      ns
     end
 
     # :call-seq:
@@ -1507,14 +1499,31 @@ module REXML
       formatter.write( self, output )
     end
 
-    private
-    def calculate_namespaces
-      if parent
-        parent.namespaces.merge(attributes.namespaces)
-      else
-        attributes.namespaces
-      end
+    # Returns namespace of the element
+    def _namespace_internal(namespaces = _calculate_namespaces) # :nodoc:
+      _namespace_lookup_internal(prefix, namespaces)
     end
+
+    # Lookup namespace for the given prefix in the context of the element
+    def _namespace_lookup_internal(prefix, namespaces = _calculate_namespaces) # :nodoc:
+      prefix = (prefix == '') ? 'xmlns' : prefix.delete_prefix("xmlns:")
+      ns = namespaces[prefix]
+      ns = '' if ns.nil? and prefix == 'xmlns'
+      ns
+    end
+
+    def _calculate_namespaces(cache_hash = nil, attlist_mappings = nil) # :nodoc:
+      return cache_hash[self] if cache_hash && cache_hash.key?(self)
+
+      attlist_mappings ||= document&.doctype&._attlist_mappings || {}
+      inherited_namespaces = parent ? parent._calculate_namespaces(cache_hash, attlist_mappings) : {}
+      attr_namespaces = attributes._calculate_namespaces(attlist_mappings)
+      namespaces = attr_namespaces.any? ? inherited_namespaces.merge(attr_namespaces) : inherited_namespaces
+      cache_hash[self] = namespaces if cache_hash
+      namespaces
+    end
+
+    private
 
     def __to_xpath_helper node
       rv = node.expanded_name.clone
@@ -2297,7 +2306,7 @@ module REXML
     #   attrs.get_attribute('nosuch')        # => nil
     #
     def get_attribute( name )
-      fetch(name, nil) || attlist_attributes&.[](name)
+      fetch(name, nil) || attlist_attributes[name]
     end
 
     # :call-seq:
@@ -2372,11 +2381,7 @@ module REXML
     #   d.root.attributes.namespaces # => {"xmlns"=>"foo", "x"=>"bar", "y"=>"twee"}
     #
     def namespaces
-      namespaces = {}
-      each_effective_attribute do |attribute|
-        namespaces[attribute.name] = attribute.value if attribute.namespace_declaration?
-      end
-      namespaces
+      _calculate_namespaces
     end
 
     # :call-seq:
@@ -2507,16 +2512,23 @@ module REXML
       end
     end
 
+    def _calculate_namespaces(attlist_mappings = nil) # :nodoc:
+      namespaces = {}
+      each_effective_attribute(attlist_mappings) do |attribute|
+        namespaces[attribute.name] = attribute.value if attribute.namespace_declaration?
+      end
+      namespaces
+    end
+
     private
 
     # Iterates over the attributes of the element and those in the DTD.
     # If the element has an attribute with the same name as one in the DTD,
     # the element's attribute takes precedence.
-    def each_effective_attribute
-      return to_enum(__method__) unless block_given?
-      attr = attlist_attributes
-      attr = attr ? attr.merge(self) : self
-      attr.each_value do |attribute|
+    def each_effective_attribute(attlist_mappings = nil)
+      return to_enum(__method__, attlist_mappings) unless block_given?
+      attributes = attlist_attributes(attlist_mappings).merge(self)
+      attributes.each_value do |attribute|
         yield attribute
       end
     end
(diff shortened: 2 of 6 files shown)


lib/rexml/element.rb
```diff
@@ -2424,19 +2447,12 @@ def prefixes
     #   d.root.attributes.namespaces # => {"xmlns"=>"foo", "x"=>"bar", "y"=>"twee"}
     #
     def namespaces
```
tompng, line 2383: This method is no longer called except from the added test.
We need to keep this if it's a public API, but if not, we can remove it (perhaps with deprecation first)

lib/rexml/element.rb
```diff
@@ -1508,12 +1500,34 @@ def write(output=$stdout, indent=-1, transitive=false, ie_hack=false)
...
+
+      inherited_namespaces = parent ? parent.send(:calculate_namespaces, cache_hash, attlist_element_namespaces) : {}
+      attlist_element_namespaces ||= document&.send(:attlist_per_element_namespaces)
+      attlist_namespaces = attlist_element_namespaces&.[](is_a?(Document) ? doctype&.name : expanded_name)
+      own_namespaces = attributes.send(:own_namespaces)
+
+      # Inherited namespaces can be overridden by attribute list declaration, and both can be overridden by its own attributes
+      namespaces = inherited_namespaces
+      namespaces = namespaces.merge(attlist_namespaces) if attlist_namespaces
+      namespaces = namespaces.merge(own_namespaces) if own_namespaces.any?
```
tompng: There was a small bug in master: precedence of attlist attributes and own attirbutes was reversed. Test is added for this

lib/rexml/doctype.rb
```diff
@@ -113,23 +113,26 @@ def node_type
...
 
     def attribute_of element, attribute
-      att_decl = find do |child|
-        child.kind_of? AttlistDecl and
-        child.element_name == element and
-        child.include? attribute
+      _attlist_mappings[element]&.[](attribute)
+    end
+
+    def _attlist_mappings
```
naitoh:
```suggestion
    def _attlist_mappings # :nodoc:
```

It is an internal method like `Element#_calculate_namespaces` and should not appear in the generated documentation.

lib/rexml/element.rb
```diff
@@ -2372,17 +2348,7 @@ def []=( name, value )
...
-        store(value.name, value)
-      elsif old_attr.kind_of? Hash
-        old_attr[value.prefix] = value
-      elsif old_attr.prefix != value.prefix
-        store value.name, {old_attr.prefix => old_attr,
-                           value.prefix    => value}
-      else
-        store value.name, value
-      end
+      store name, value
```
naitoh, line 2355: - before (expect)
```ruby
> doc = REXML::Document.new("<a/>")
> doc.root.attributes['x'] = REXML::Attribute.new('y:z', 'v')
> doc.root # => <a y:z='v'/>
> doc.root.attributes['y:z'] # => "v"
> doc.root.attributes['x'] # => nil
```

- after (this PR)
```ruby
> doc = REXML::Document.new("<a/>")
> doc.root.attributes['x'] = REXML::Attribute.new('y:z', 'v')
> doc.root # => <a y:z='v'/>
> doc.root.attributes['y:z'] # => nil
> doc.root.attributes['x'] # => "v"
```

```suggestion
      store value.expanded_name, value
```

lib/rexml/xpath_parser.rb
```diff
@@ -85,14 +88,8 @@ def parse path, node
         node = node.first
       end
 
-      document = node.document
-      if document
-        document.__send__(:enable_cache) do
```
naitoh, line 90: If `Document#enable_cache` and `namespaces_cache` are dead code, let's remove them.

https://github.com/ruby/rexml/blob/24035a201a3bff2984fb81ceb06e26a9f9f7c237/lib/rexml/document.rb#L452-L464

test/xpath/test_base.rb
```diff
@@ -1312,16 +1312,6 @@ def test_namespaces_0
...
-      assert_equal("<b/>", XPath.first(doc, "//b[namespace-uri()='1']").to_s)
-      assert_nil(XPath.first(doc, "//b[namespace-uri()='']"))
-
-      doc.root.delete_namespace
-      assert_nil(XPath.first(doc, "//b[namespace-uri()='1']"))
-      assert_equal("<b/>", XPath.first(doc, "//b[namespace-uri()='']").to_s)
-    end
-
     def test_ticket_71
       doc = Document.new(%Q{<root xmlns:ns1="xyz" xmlns:ns2="123"><element ns1:attrname="foo" ns2:attrname="bar"/></root>})
```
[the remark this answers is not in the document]
naitoh: @tompng 
I think the behavior that `test_namespaces_cache` was verifying is still worth preserving, but do we need to remove it?

lib/rexml/xpath_parser.rb
```diff
@@ -85,14 +88,8 @@ def parse path, node
...
-          match( path_stack, node )
-        end
-      else
-        match( path_stack, node )
-      end
+      @document = node.document
```
naitoh: - before (expect)
```ruby
> doc = REXML::Document.new '<!DOCTYPE a [<!ATTLIST b xmlns:p CDATA "urn:x">]><a><b><p:c/></b></a>'
> parser = REXML::XPathParser.new
> parser.predicate("//p:c", doc) #=> [<p:c/>]
```

- after (this PR)
```ruby
> doc = REXML::Document.new '<!DOCTYPE a [<!ATTLIST b xmlns:p CDATA "urn:x">]><a><b><p:c/></b></a>'
> parser = REXML::XPathParser.new
> parser.predicate("//p:c", doc) # => []
```

Although `@document` is set in `parse`, it is not set in `predicate` or `get_first`, so `@attlist_mappings` becomes `{}` in those cases, and the xmlns declaration derived from `ATTLIST` does not take effect.

lib/rexml/element.rb
```diff
@@ -2562,17 +2490,58 @@ def delete_all( name )
...
+    end
+
+    private
+
+    # Iterates over the attributes of the element and those in the DTD.
+    # If the element has an attribute with the same name as one in the DTD,
+    # the element's attribute takes precedence.
+    def each_effective_attribute(attlist_mappings = nil)
+      return to_enum(__method__, attlist_mappings) unless block_given?
+      attributes = attlist_attributes(attlist_mappings).merge(self)
```
naitoh: If there is no `ATTLIST` declaration, I want to avoid duplicating the attribute hash.

```suggestion
      attlist = attlist_attributes(attlist_mappings)
      attributes = attlist.empty? ? self : attlist.merge(self)
```

lib/rexml/element.rb
```diff
@@ -2562,17 +2490,58 @@ def delete_all( name )
...
     def get_attribute_ns(namespace, name)
-      result = nil
-      each_attribute() { |attribute|
-        if name == attribute.name &&
-          namespace == attribute.namespace() &&
-          ( !namespace.empty? || !attribute.fully_expanded_name.index(':') )
-          # foo will match xmlns:foo, but only if foo isn't also an attribute
-          result = attribute if !result or !namespace.empty? or
-                                !attribute.fully_expanded_name.index(':')
+      each_effective_attribute.find do |attribute|
```
naitoh, line 2497: Should explicit attributes take precedence over the default values in ATTLIST?

 
- before
```ruby
> doc = REXML::Document.new '<!DOCTYPE a [<!ATTLIST a x:v CDATA "default">]><a xmlns:x="urn:n" xmlns:y="urn:n" y:v="explicit"/>'
> doc.root.attributes.get_attribute_ns('urn:n', 'v').value
=> "explicit"
```

- after (this PR)
```ruby
> doc = REXML::Document.new '<!DOCTYPE a [<!ATTLIST a x:v CDATA "default">]><a xmlns:x="urn:n" xmlns:y="urn:n" y:v="explicit"/>'
> doc.root.attributes.get_attribute_ns('urn:n', 'v').value
=> "default"
```

tompng: In that case, both `x:v` and `y:v` have `name=='v'` and `namespace=='urn:n'`, but there shouldn't be a duplicated attributes with the same namespace-uri and name.
The sample below is in a same situation:
```ruby
xml = '<a xmlns:x="urn:n" xmlns:y="urn:n" x:v="foo" y:v="bar"/>'
doc = REXML::Document.new xml, ignore_namespace_conflict: true # note: this option doesn't exist
doc.root.attributes.get_attribute_ns('urn:n', 'v').value
#=> "foo" or "bar", non-deterministic
```

test/xpath/test_base.rb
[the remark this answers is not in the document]
tompng: We're not caching something into an instance variable of document anymore. In my opinion, this test was valuable only because something is cached to document's instance variable.
It may be worth adding this to detect regression when we re-introducing document-level cache, but I think organizing and tidy up outdated tests are also worth.

lib/rexml/xpath_parser.rb
tompng: removed

lib/rexml/element.rb
tompng: name (as a key) is given as a first argument, I think `store name, value` is readable.
Fixed by overriding `name` with `value.expanded_name` as:
```
+ if value.kind_of? Attribute
+   # Use attribute's expanded name to avoid inconsistency.
+   # TODO: Explicitly document that inconsistent names are allowed, or raise/warn about inconsistency.
+   name = value.expanded_name
+ else
- unless value.kind_of? Attribute
    ...
  end
```
Applied to #335 and rebased.

lib/rexml/xpath_parser.rb
tompng: Fixed. But I can't confidently say that it works because `predicate` is not tested at all, and `first` (called from `get_first`) is an incomplete function.
```ruby
# FIXME: This method is incomplete!
def first( path_stack, node )
  ...
end
```
Maybe it's worth deleting them in a separate pull request.

lib/rexml/element.rb
naitoh: OK, I see.

test/xpath/test_base.rb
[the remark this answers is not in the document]
naitoh: OK, I see.

lib/rexml/attribute.rb
```diff
@@ -205,6 +205,15 @@ def xpath
     def document
       @element&.document
     end
+
+    # Returns true if this attribute is a namespace declaration, false otherwise.
+    #   "foo" => false
+    #   "xmlns" => true
+    #   "xmlns:foo" => true
+    #   "foo:xmlns" => false
+    def namespace_declaration?
```
naitoh, line 214: Could you add a test?

tompng: Applied in 28979eb445ea918a46331e32bcd71ba867ae4f9c

lib/rexml/xpath_parser.rb
```diff
@@ -152,6 +149,7 @@ def first( path_stack, node )
 
 
     def match(path_stack, node)
+      @document = node.document
```
naitoh, line 152: When reusing an XPathParser, old cache entries remain. 

* before (expect)
```ruby
> doc = REXML::Document.new("<a xmlns='u1'><b/></a>")
> parser = REXML::XPathParser.new
> parser.namespaces = {"p" => "u1"}
> parser.parse("//p:b", doc)
=> [<b/>]
> doc.root
=> <a xmlns='u1'> ... </>
> doc.root.delete_namespace
=> <a> ... </>
> doc.root
=> <a> ... </>
> parser.parse("//p:b", doc)
=> []
```
* after (this PR)
```ruby
> doc = REXML::Document.new("<a xmlns='u1'><b/></a>")
> parser = REXML::XPathParser.new
> parser.namespaces = {"p" => "u1"}
> parser.parse("//p:b", doc)
=> [<b/>]
> doc.root.delete_namespace
=> <a> ... </>
> doc.root
=> <a> ... </>
> parser.parse("//p:b", doc)
=> [<b/>]
```


```suggestion
      @document = node.document
      @attlist_mappings = nil
      @element_namespaces_cache = {}
```

lib/rexml/xpath_parser.rb
naitoh: > Fixed. 

Thanks

> Maybe it's worth deleting them in a separate pull request.

We might want to consider deleting it.

lib/rexml/xpath_parser.rb
naitoh: @tompng

tompng: Old values remain, but it doesn't cause a problem as long as XPathParser is one-off or internal use only.
For example, `XPath::match` creates an `XPathParser` instance, use it once, and don't reuse it. Module function creates an instance, use once, and discard it, is a frequently used pattern, I think.

To maintain reusability, I think it's better to separate initialize into variable-reset-method and other setups like this:
```ruby
def initialize(strict: false)
  # Setup-once values: @debug, @strict
  # Writable values: @namespaces, @variables, @functions

  # Other variables to be reset before reuse
  # Cache: @attlist_mappings, @element_namespaces_cache, @document
  # State: @nest
  # Values that cannot be definitively classified as stateless/reusable: @parser
  reset_other_instance_variables
end

def match(path_stack, node)
  reset_other_instance_variables
  ...
end
```
However, the simplest way is to make/leave that XPathParser is one-off and/or internal. And I think it is already documented as internal-use. What do you think?
https://github.com/ruby/rexml/blob/73aa07a9e4fb3641023bdc1aa6bf3b0df84303f1/lib/rexml/xpath_parser.rb#L50-L54

naitoh: OK.
This is already for internal use.
No changes are needed.
Thank you!

https://ruby.github.io/rexml/REXML/XPathParser.html
https://docs.ruby-lang.org/ja/latest/class/REXML=3a=3aParsers=3a=3aXPathParser.html

naitoh (approved):
I think that's a great improvement.
Thank you.

PR #335, opened 2026-06-15
Simplify element.attributes structure

tompng:
Basically refactor Attributes, and fixes some bugs found while doing it.

## Structure of Attributes, a subclass of Hash
For XML: `<element name1="1" ns1:name2="2" ns2:name2="3" ns1:name3="4"/>`, the structure was:
```ruby
{
  name1 => attr1
  name2 => { ns1 => attr2, ns2 => attr3 }
  name3 => att4
}
```
This structure is useful for fuzzy name search, and fuzzy name deletion, but we don't need this complicated structure because fuzzy search should be removed, and fuzzy deletion is not used from REXML (and it wasn't working at all)

I think the mental model of namespaced attribute in Document Object Model is something like this.
```ruby
{ default: { name1: attr1 }, ns1: { name2: attr1, name3: attr4}, ns2: { name2: attr2 } }
# or simply
{ "name1": attr1, "ns1:name2": attr2, "ns2: name2": attr3, "ns1:name3": attr4 }
```
This PR changes it to the latter, simple one.

## Stop fuzzy get_attribute/get_attribute_ns search

`Attributes#get_attribute` does fuzzy search only on its own attribute, not on attributes defined in ATTLIST. `Attributes#get_attribute_ns` tries fuzzy search but fails because of a bug.
In Document Object Model, get_attribute and get_attribute_ns should perform strictly no-prefix attribute if prefix/namespace is not given.

## ATTLIST lookup
There's many redundant duplicated code for ATTLIST lookup.
`Attributes#get_attribute_ns` forgot to check ATTLIST (bug).
Introduce `each_effective_attribute` that iterates attributes defined in ATTLIST and element's own attributes

## Fixed bugs
- get_attribute: Fuzzy search shouldn't be performed
- get_attribute_ns: forgot to lookup ATTLIST
- delete_all: Unlike the name that indicates fuzzy deletion, it deletes only one attribute, and returns wrong attrribute which is a different one than actually deleted when prefix is given.
- namespace declaration check: `foo:xmnls` was wrongly considered namespace declaration
- And perhaps more which I haven't found yet

### DIFF
--- a/lib/rexml/attribute.rb
+++ b/lib/rexml/attribute.rb
@@ -180,7 +180,7 @@ module REXML
     #
     # This method is usually not called directly.
     def remove
-      @element.attributes.delete self.name unless @element.nil?
+      @element.attributes.delete self unless @element.nil?
     end
 
     # Writes this attribute (EG, puts 'key="value"' to the output)
@@ -204,6 +204,15 @@ module REXML
 
     def document
       @element&.document
+    end
+
+    # Returns true if this attribute is a namespace declaration, false otherwise.
+    #   "foo" => false
+    #   "xmlns" => true
+    #   "xmlns:foo" => true
+    #   "foo:xmlns" => false
+    def namespace_declaration?
+      prefix == 'xmlns' || (prefix.empty? && name == 'xmlns')
     end
   end
 end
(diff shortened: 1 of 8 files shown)


naitoh (approved):
I think this PR is reasonable corrections.
Thank you!

PR #339, opened 2026-06-20
XPath optimize preparation for positional predicates that uses `last()`

tompng:
Expand optimizable simple positional queries to support `[last()]` `[last()-N]` and `[position() <=> last()-N]`.
This will open the door for future optimization.

Simple positional predicates that may be optimized are represented as:
```
Before: [:==, 1], [:<, 2], [:>, 3]
↓
After: [:index_eq, 0], [:index_lt, 1], [:index_gt, 2]
Added: [:reverse_index_eq, 0], [:reverse_index_lt, 2], [:reverse_index_gt, 4]
```

# Example
Example optimization of preceding-sibling/following-sibling (not included in this PR)
```ruby
def preceding_following_sibling(nodeset, tester, selector, reverse:)
  ...
  when :reverse_index_eq, :reverse_index_lt, :reverse_index_gt
    nodeset.group_by(&:parent).flat_map do |parent, sibling_nodes|
      anchors = Set.new.compare_by_identity
      sibling_nodes.each {|sibling| anchors << sibling }
      children = parent.children
      children = children.reverse if reverse

      # Different anchor node gives the same reverse-index. We only need to check with the first anchor
      followings = children.drop_while {|child| !anchors.include?(child) }.drop(1)
      candidates = followings.select(&tester).reverse
      case operator
      when :reverse_index_eq
        matched = candidates[value] if value >= 0
        matched ? [matched] : []
      when :reverse_index_lt
        value >= 0 ? candidates[0...value] : []
      when :reverse_index_gt
        value >= 0 ? candidates.drop(value + 1) : candidates
      end
    end
  when :nodesets
    ...
  end
end
```

```ruby
doc = REXML::Document.new('<root>'+('<a>'+'<b/><c/>'*500+'</a>')*2+'</root>')
REXML::XPath.match(doc, '//c/preceding-sibling::*[position()<last()-500]').size
# processing time:  9.740037s → 0.059448s
# => 996
```

### DIFF
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -385,31 +385,63 @@ module REXML
       expr.any? {|part| calls_position_dependent_function?(part) }
     end
 
-    # Detects simple position-based predicates that can be optimized in axis scanning, such as [1], [position()=1], [position() < 2], [position() > 3]
-    # Returns operators and values such as [:==, 1], [:<, 2], [:>, 3]
+    # Detects simple position-based predicates that can be optimized in axis scanning, such as [1], [position()=1], [position() < 2], [last()-3], etc.
+    # Returns operators and values such as [:index_eq, 0], [:index_lt, 1], [:index_gt, 2], [:reverse_index_eq, 0], [:reverse_index_lt, 1], [:reverse_index_gt, 2]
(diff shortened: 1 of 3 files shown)


naitoh (approved):
Thanks!

PR #341, opened 2026-06-27
Optimize preceding-sibling, following-sibling, following, and descendant axes with an unified single mechanism

### LINKED ISSUE #331: XMLDecl mixed in xpath `descendant::node()` match result
XMLDecl should be excluded
```ruby
# Expected
Nokogiri::XML.parse('<?xml version="1.0"?><root></root>').xpath("descendant::node()")
# => [#<Nokogiri::XML::Element:0xb38 name="root">]

# Actual
REXML::XPath.match REXML::Document.new('<?xml version="1.0"?><root></root>'), "descendant::node()"
# => [<?xml ... ?>, <root/>]
REXML::VERSION
# => "3.4.4"
```

tompng:
GitHub: Fix GH-331

Unify the positional-predicate handling of the `descendant`, `descendant-or-self`,
`following`, `following-sibling`, and `preceding-sibling` axes into a single
event-stream mechanism (`sequence_positional_scan`).

## Detail

Consider xpath like this: `anchor/axis::*[test-predicate][simple-positional-predicate]`.
Some xpath axes (descendant, descendant-or-self, following, preceding-sibling, following-sibling) have same common structure:
- Anchor start point, anchor end point, and nodes that passed test predicate lines up in a single sequence
- Anchor start and end are nested and doesn't cross over

Example of descendant:
```
anchor1-start
  node1
  node2
  anchor2-start
    node3
    node4
  anchor2-end
  node5
  anchor3-start
    node6
  anchor3-end
  node7
anchor1-end

# Nodesets: [
#   [node1, node2, node3, node4, node5, node6, node7], # from anchor1
#   [node3, node4], # from anchor2
#   [node6] # from anchor3
# ]
```

The above sequence and anchor ranges can be represented as event stream like this:
```ruby
[:push, node1, node2, :push, node3, node4, :pop, node5, :push, node6, :pop, node7, :pop]
```
Events are:
- `:push`: Add a new anchor point
- `:pop`: Remove last anchor point
- `node`: Add a node that passed test predicate

Axis scanner of `following`, `descendant`, `descendant-or-self`, `preceding-sibling` and `following-sibling` will construct an event stream and passes it to `sequence_positional_scan` which implements positional predicate optimization.
Optimization logic in `sequence_positional_scan` is basically the same as the one implemented in `preceding/following-sibling` before.

## Unchanged axes (preceding, ancestor)

`preceding` and `ancestor` axes are intentionally left on the existing path. Unlike `following` axis, anchor-exclusion semantics of `preceding` axis don't fit the nested push/pop model.

## Benchmark

XPath in this benchmark is specially crafted to precisely measure axis scan without noise.
Using `*` and `count()` will avoid `O(n^2)` sort and namespace lookup which may be fixed in a near future.

```ruby
xml = '<a>'*200+'<a/>'*200+'</a>'*200
xpaths = [
  "count(//*/descendant::*[position()<10])",
  "count(//*/preceding-sibling::*[position()<10])",
  "count(//*/following-sibling::*[position()<10])",
  "count(//*/following::*[position()<10])",
]
xpaths.each do |xpath|
  puts xpath
  t=Time.now; p [REXML::XPath.match(doc, xpath), Time.now-t]
  doc = Nokogiri::XML.parse(xml)
  t=Time.now; p [doc.xpath(xpath), Time.now-t]
end
```

| scenario | REXML(master) | REXML(this PR) | Nokogiri |
| --- | --- | --- | --- |
| `descendant::*[position()<10]` | 0.032287 sec | 0.002089 sec | 0.010074 sec |
| `preceding-sibling::*[position()<10]` | 0.001075 sec | 0.001239 sec | 0.00265 sec |
| `following-sibling::*[position()<10]` | 0.000892 sec | 0.001041 sec | 0.002627 sec |
| `following::*[position()<10]` | 0.110464 sec | 0.000942 sec | 0.002706 sec |

In this example, XPath match is faster than Nokogiri, mainly because nokogiri doesn't optimize `[position()<N]`.

### DIFF
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -467,57 +467,18 @@ module REXML
 
     def preceding_following_sibling(nodeset, tester, selector, reverse:)
       nodeset = nodeset.select {|node| node.respond_to?(:parent) && node.parent }
-      operator, value = selector
-      case operator
-      when :uniq
-        nodeset.group_by(&:parent).flat_map do |parent, sibling_nodes|
-          sets = Set.new.compare_by_identity
-          sibling_nodes.each {|sibling| sets << sibling }
-          children = parent.children
-          children = children.reverse if reverse
-          children.drop_while {|child| !sets.include?(child) }.drop(1)
-        end.select(&tester)
-      when :index_eq, :index_lt, :index_gt
-        nodeset.group_by(&:parent).flat_map do |parent, sibling_nodes|
-          anchors = Set.new.compare_by_identity
-          sibling_nodes.each {|sibling| anchors << sibling }
-          children = parent.children
-          children = children.reverse if reverse
-          followings = children.drop_while {|child| !anchors.include?(child) }.drop(1)
-          anchor_indexes = Set[0]
-          last_anchor = 0
-          index = 0
-          matched = []
-          followings.each do |node|
-            if tester.call(node)
-              case operator
-              when :index_eq
-                # anchor_indexes only contain values smaller or equal to `index`,
-                # so value < 0 case doesn't accidentally match any node.
-                matched << node if anchor_indexes.include?(index - value)
-              when :index_lt
-                # Position from the last anchor will be the minimum possible position for the node
-                matched << node if index - last_anchor < value
-              when :index_gt
-                # Position from the first anchor(==0) will be the maximum possible position for the node
-                matched << node if index > value
-              end
-              index += 1
-            end
-            if anchors.include?(node)
-              anchor_indexes << index
-              last_anchor = index
-            end
-          end
-          matched
+      nodeset.group_by(&:parent).flat_map do |parent, sibling_nodes|
+        anchors = Set.new.compare_by_identity.replace(sibling_nodes)
+        children = parent.children
+        children = children.reverse if reverse
+        followings = children.drop_while {|child| !anchors.include?(child) }.drop(1)
+        events = [:push]
+        followings.each do |node|
+          events << node if tester.call(node)
+          events << :push if anchors.include?(node)
         end
-      else # Slow path for :nodesets, :reverse_index_eq, :reverse_index_lt, :reverse_index_gt
-        nodesets = nodeset.map do |node|
-          parent = node.parent
-          index = parent.children.index(node)
-          reverse ? parent.children[0...index].reverse : parent.children[index + 1..-1]
-        end
-        non_optimized_nodesets_select(nodesets, tester, selector)
+        anchors.size.times { events << :pop }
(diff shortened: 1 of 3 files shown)


naitoh:
@tompng 
Thank you.
I think this is a great improvement.

However, please check the following.

1. This PR is in conflict.
2. It appears there is a performance degradation of about 9% in `//c`.

```sh
$ benchmark-driver benchmark/xpath_latest.yaml
                               before(master)       after  before(master)(YJIT)  after(YJIT) 
REXML::XPath.match(doc, '//c')         93.025      84.225               126.617      116.110 i/s -     100.000 times in 1.074978s 1.187298s 0.789785s 0.861253s

Comparison:
             REXML::XPath.match(doc, '//c')
          before(master)(YJIT):       126.6 i/s 
                   after(YJIT):       116.1 i/s - 1.09x  slower
                before(master):        93.0 i/s - 1.36x  slower
                         after:        84.2 i/s - 1.50x  slower
```

### Benchmark Code

```
$ git checkout master
$ rake build
$ gem install pkg/rexml-3.4.5.gem 
```

```YAML
benchmark/xpath_latest.yaml
loop_count: 100
contexts:
  - name: before(master)
    gems:
      rexml: 3.4.5
    require: false
    prelude: require 'rexml'
  - name: after
    prelude: |
      require 'rexml'
  - name: before(master)(YJIT)
    gems:
      rexml: 3.4.5
    require: false
    prelude: |
      require 'rexml'
      RubyVM::YJIT.enable
  - name: after(YJIT)
    prelude: |
      require 'rexml'
      RubyVM::YJIT.enable

prelude: |
  require 'rexml/document'

  xml = "<r>" + "<g><h>#{'<b><c/></b>' * 50}</h></g>" * 40 + "</r>"

benchmark:
  "REXML::XPath.match(doc, '//c')" : REXML::XPath.match(doc, "//c")
```

tompng:
Conflict: rebased.
This removes each `selector = :uniq` optimization and replace it to a generic one. Special optimization is always faster than generic optimization, but will introduce complexity.
I think simplicity/maintainability is important, and compared to that, +9% is a small change. Also note that this is already incomparably faster than the latest released version due to a better time complexity.
How about reconsider adding special optimization after this pull request, and compare if it's speed up is worth adding complexity for each axis.

naitoh:
> Conflict: rebased. 

Thanks.

> How about reconsider adding special optimization after this pull request, and compare if it's speed up is worth adding complexity for each axis.

OK.
I see.

naitoh (approved):
I was able to understand the new algorithm.
The steps are well-organized, and I think it’s easier to understand than before.
Thank you very much.

PR #342, opened 2026-06-28
Fix XPath variable handling and invalid value contamination

### LINKED ISSUE #15: XPath: fix default node detection in functions
We need similar fix with #13 not only `string` but also `namespace_uri`, `name`, `normalize_space` and `number`.

tompng:
Fixes #15

Fix nil and other invalid value contamination. Nil can be injected through variables and through id function.
Unimplemented function `id()` is fixed to return `[]` instead of `nil`.
Applies predicates/path evaluation if it exists after variable.
Add variable validation (TypeError if invalid).

Variable existence check (NameError if undefined variable exist in xpath even if it is not referenced) is excluded from this PR, and just fallbacks to "" for now.

Minor behavior changes for an invalid XPath match:
```ruby
doc = REXML::Document.new("<root/>")
REXML::XPath.match(doc, '($x)[1<2]', {}, {'x'=>42})
#=> [42] → []
REXML::XPath.match(doc, '$x[1<2]', {}, {'x'=>42})
#=> [42] → []
```
It may raise TypeError (because `$x` wasn't evaluated to a nodeset), though, it shoudn't return `[42]`

## Note
Although nodeset as a variable has been accepted before, the code added in pull request explicitly permits it.
Nokogiri only accepts string value as a variable, so there's an option to limit the value types here.
Accepting nodeset may cause a problem if the passed nodes doesn't belong to a single document root. (or  just consider such case as unsupported)

### DIFF
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -72,7 +72,8 @@ module REXML
       @namespaces = namespaces
     end
 
-    def variables=( vars={} )
+    def variables=(vars)
+      vars = vars.transform_values { |v| coerce_variable(v) }
       @functions.variables = vars
       @variables = vars
     end
@@ -244,7 +245,8 @@ module REXML
           end
         when :variable
           var_name = path_stack.shift
-          value = coerce_variable(@variables[var_name])
+          value = @variables[var_name]
+          raise NameError, "Undefined variable: $#{var_name}" if value.nil?
           return value if path_stack.empty?
 
           nodeset = apply_remaining_predicates(path_stack, value)
@@ -346,18 +348,23 @@ module REXML
       end
     end
 
-    # Coerces a variable value to a type that can be used in XPath expressions.
-    # TODO: Decide whether REXML should warn, raise, or ignore when a variable value is invalid.
+    # Validates and coerces a variable value to a type that can be used in XPath expressions.
     def coerce_variable(value)
       case value
       when REXML::Node
         [value]
       when Array
-        value.grep(REXML::Node).uniq
+        unless value.all?(REXML::Node)
+          raise TypeError, 'Array variable must contain only REXML::Node objects'
+        end
+        value.uniq
       when Numeric, String, true, false
         value
-      else
+      when nil
+        # Convert to an empty string for backward compatibility
         ""
+      else
+        raise TypeError, "Unsupported variable type: #{value.class}"
       end
     end
 
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -107,7 +107,7 @@ module REXML
     end
 
     def []=( variable_name, value )
-      @variables[ variable_name ] = value
+      @variables[variable_name] = coerce_variable(value)
     end
 
 
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -347,6 +347,8 @@ module REXML
     # TODO: Decide whether REXML should warn, raise, or ignore when a variable value is invalid.
     def coerce_variable(value)
       case value
+      when REXML::Node
+        [value]
       when Array
         value.grep(REXML::Node).uniq
       when Numeric, String, true, false
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -244,8 +244,10 @@ module REXML
           end
         when :variable
           var_name = path_stack.shift
-          return @variables[var_name]
+          value = coerce_variable(@variables[var_name])
+          return value if path_stack.empty?
 
+          nodeset = apply_remaining_predicates(path_stack, value)
         when :eq, :neq, :lt, :lteq, :gt, :gteq
           left = expr( path_stack.shift, nodeset.dup, context )
           right = expr( path_stack.shift, nodeset.dup, context )
@@ -319,15 +321,9 @@ module REXML
         when :group
           sub_expression = path_stack.shift
           result = expr(sub_expression, nodeset, context)
-          if result.is_a?(Array)
-            # If result is a nodeset, apply following predicates
-            path_stack.unshift(:node)
-            nodeset = step(path_stack) do
-              [:iterate_nodesets, [XPathParser.sort(result)]]
-            end
-          else
-            return result
-          end
+          return result if path_stack.empty?
+
+          nodeset = apply_remaining_predicates(path_stack, result)
         else
           raise "[BUG] Unexpected path: <#{op.inspect}>: <#{path_stack.inspect}>"
         end
@@ -335,6 +331,29 @@ module REXML
       return nodeset
     ensure
       leave(:expr, path_stack, nodeset) if @debug
+    end
+
+    def apply_remaining_predicates(path_stack, value)
+      # If evaluated value is not a nodeset, treat it as an empty nodeset.
+      # TODO: Decide whether REXML should raise type error or keep this behavior.
+      value = [] unless value.is_a?(Array)
+      path_stack.unshift(:node)
+      step(path_stack) do
+        [:iterate_nodesets, [XPathParser.sort(value)]]
+      end
+    end
+
+    # Coerces a variable value to a type that can be used in XPath expressions.
+    # TODO: Decide whether REXML should warn, raise, or ignore when a variable value is invalid.
+    def coerce_variable(value)
+      case value
+      when Array
+        value.grep(REXML::Node).uniq
+      when Numeric, String, true, false
+        value
+      else
+        ""
+      end
     end
 
     # Determines if a predicate expression is dependent on the position of nodes.
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -338,8 +338,6 @@ module REXML
     end
 
     def apply_remaining_predicates(path_stack, value)
-      # If evaluated value is not a nodeset, treat it as an empty nodeset.
-      # TODO: Decide whether REXML should raise type error or keep this behavior.
       value = [] unless value.is_a?(Array)
       path_stack.unshift(:node)
       step(path_stack) do
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -317,7 +317,10 @@ module REXML
             expr(arg, nodeset, target_context)
           end
           @functions.context = target_context
-          return @functions.send(func_name, *args)
+          result = @functions.send(func_name, *args)
+          return result if path_stack.empty?
+
+          nodeset = apply_remaining_predicates(path_stack, result)
         when :group
           sub_expression = path_stack.shift
           result = expr(sub_expression, nodeset, context)
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -245,8 +245,7 @@ module REXML
           end
         when :variable
           var_name = path_stack.shift
-          value = @variables[var_name]
-          raise NameError, "Undefined variable: $#{var_name}" if value.nil?
+          value = @variables.key?(var_name) ? @variables[var_name] : ""
           return value if path_stack.empty?
 
           nodeset = apply_remaining_predicates(path_stack, value)
--- a/lib/rexml/functions.rb
+++ b/lib/rexml/functions.rb
@@ -57,6 +57,7 @@ module REXML
     # Since REXML is non-validating, this method is not implemented as it
     # requires a DTD
     def id( object )
+      []
     end
 
     def local_name(node_set=nil)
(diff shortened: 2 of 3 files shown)


lib/rexml/xpath_parser.rb
```diff
@@ -337,6 +333,29 @@ def expr( path_stack, nodeset, context=nil )
...
+    def coerce_variable(value)
+      case value
+      when Array
+        value.grep(REXML::Node).uniq
+      when Numeric, String, true, false
+        value
+      else
+        ""
+      end
+    end
```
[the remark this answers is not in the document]
naitoh: @tompng 
I think we should implement this suggestion from Copilot. 

* before (expect)
```ruby
> doc = REXML::Document.new("<a><b><c/></b><d><e/></d></a>")
> a, b, c, d, e = REXML::XPath.match(doc, '//*')
> REXML::XPath.match(doc, '$x', nil, { 'x' => b })        #=> [<b> ... </>]
> REXML::XPath.match(doc, '$x/*', nil, { 'x' => b })      #=> Garbage component exists at the end: </*>: <$x/*> (REXML::ParseException)
> REXML::XPath.match(doc, 'count($x)', nil, { 'x' => b }) #=> [1]
```

I think `REXML::XPath.match(doc, “$x/*”, nil, { “x” => b })` is valid for `[<c> ... </>]`.

* after(this PR)
```ruby
> doc = REXML::Document.new("<a><b><c/></b><d><e/></d></a>")
> a, b, c, d, e = REXML::XPath.match(doc, '//*')
> REXML::XPath.match(doc, '$x', nil, { 'x' => b })        #=> [""]
> REXML::XPath.match(doc, '$x/*', nil, { 'x' => b })      #=> []
> REXML::XPath.match(doc, 'count($x)', nil, { 'x' => b }) #=> [0]
```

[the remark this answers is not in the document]
tompng: Before, a single node, (not a nodeset) was silently passed, and may cause unintentional behavior. Perhaps it worked in some case, but it wasn't valid.
Also note that `$x/subpath` and `$x[predicate]` was not possible before. It's unable to verify regression in a practical case.
I think it's ok to convert a single node to a single nodeset, as a new feature. It'll be useful, but as a refactor/bugfix, I explicitly dropped from this pull request.
If you want to change it in this PR, I can add it.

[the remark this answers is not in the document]
tompng: I reconsidered, I think supporting it is reasonable. Applied in a31f5cedb5b0680e4fa69c6ead7863d7e9cb4906

lib/rexml/xpath_parser.rb
```diff
@@ -337,6 +333,31 @@ def expr( path_stack, nodeset, context=nil )
       leave(:expr, path_stack, nodeset) if @debug
     end
 
+    def apply_remaining_predicates(path_stack, value)
+      # If evaluated value is not a nodeset, treat it as an empty nodeset.
+      # TODO: Decide whether REXML should raise type error or keep this behavior.
```
naitoh: > If evaluated value is not a nodeset, treat it as an empty nodeset.
>  TODO: Decide whether REXML should raise type error or keep this behavior.

According to the XPath specification, anything other than a node set results in an error.

https://www.w3.org/TR/1999/REC-xpath-19991116/#node-sets

>  It is an error if the expression to be filtered does not evaluate to a node-set. 

>  It is an error if the expression does not evaluate to a node-set. 

I think it would be better to raise an error, as that would make the processing clearer. 
What do you think? @kou @tompng

lib/rexml/xpath_parser.rb
```diff
@@ -337,6 +333,31 @@ def expr( path_stack, nodeset, context=nil )
...
+      value = [] unless value.is_a?(Array)
+      path_stack.unshift(:node)
+      step(path_stack) do
+        [:iterate_nodesets, [XPathParser.sort(value)]]
+      end
+    end
+
+    # Coerces a variable value to a type that can be used in XPath expressions.
+    # TODO: Decide whether REXML should warn, raise, or ignore when a variable value is invalid.
```
naitoh: >    Coerces a variable value to a type that can be used in XPath expressions.
>    TODO: Decide whether REXML should warn, raise, or ignore when a variable value is invalid.

According to the XPath specification, if a variable name is not bound to any value, an error occurs.

https://www.w3.org/TR/1999/REC-xpath-19991116/#section-Basics

> A VariableReference evaluates to the value to which the variable name is bound in the set of variable bindings in the context.

I think it would be better to raise an error, as that would make the processing clearer.
What do you think? @kou @tompng

tompng: I think referencing unbound variable should raise error. Concern: when to do that, and what kind of error should it raise.
```ruby
when :variable
  var_name = path_stack.shift
  # Handle unbound variables
  raise unless @variables.key?(var_name)

  # Handle conversion and validation of variable value
  # @variables[var_name]==nil case is handled here
  value = coerce_variable(@variables[var_name])
  ...
```

Even if we reject unbound variables, this TODO comment for bound variables (reject bound invalid-value variable or convert it to a fallback value "") still remains.

naitoh: > Concern: when to do that, and what kind of error should it raise.

I think it would be better to check immediately after parsing and before evaluation begins, rather than raising an exception when the `variable` statement is evaluated.
I think `NameError` is the appropriate error type to raise.

Reasons:

- If we raise only when the reference is actually evaluated, the error becomes document-dependent. For example, `//foo[@bar=$typo]` does not raise when no foo reaches the predicate, and or/and short-circuits can also skip the reference. So the typo silently returns `[]` in exactly the case we most want to detect.

> Even if we reject unbound variables, this TODO comment for bound variables (reject bound invalid-value variable or convert it to a fallback value "") still remains.

I think the preferred design is to fall back to `""` for compatibility purposes only when the value is nil, and to reject the input in all other cases.
In this case, I think `TypeError` is the appropriate type of error to raise.

- https://github.com/ruby/rexml/pull/13

lib/rexml/xpath_parser.rb
tompng: I think it's better to align with the spec, but it requires much more changes
In my opinion, it must out of scope of this PR, must be done in a separated PR on top of this and on #343
For example, these xpath that requires requires intermediate value to be a nodeset should be changed to raise error.
```
# Function argument that requires nodeset
local-name(1)
sum("string")
count(1 < 2)

# Path after function
string(/a)[predicate]
string(/a)/path

# Path after non-nodeset variable
$x[predicate]
$x/path

# path after literal
"string"[predicate]
42/path

# path after neg
-number(/a)[predicate]
-number(/a)/path
```

lib/rexml/xpath_parser.rb
tompng: > I think it would be better to check immediately after parsing and before evaluation

It makes sense. Applied in 7cdffee3cc08537f5a6ecfe40ed3b58f60e79b57
The added check is worth enabled for normal use, so it is applied even if strict is false (default)

lib/rexml/xpath_parser.rb
```diff
@@ -244,8 +245,11 @@ def expr( path_stack, nodeset, context=nil )
           end
         when :variable
           var_name = path_stack.shift
-          return @variables[var_name]
+          value = @variables[var_name]
+          raise NameError, "Undefined variable: $#{var_name}" unless value
+          return value if path_stack.empty?
```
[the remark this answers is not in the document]
tompng: Fixed and force pushed

lib/rexml/xpath_parser.rb
```diff
@@ -337,6 +338,36 @@ def expr( path_stack, nodeset, context=nil )
...
+    # Validates and coerces a variable value to a type that can be used in XPath expressions.
+    def coerce_variable(value)
+      case value
+      when REXML::Node
+        [value]
+      when Array
+        unless value.all?(REXML::Node)
+          raise TypeError, 'Array variable must contain only REXML::Node objects'
+        end
+        value.uniq
```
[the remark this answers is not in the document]
tompng: Ensured to work correctly with test added in d19849a59c1d1f8461acd25d3f1f0da1546329e5

lib/rexml/xpath_parser.rb
naitoh: OK.
I see.

lib/rexml/xpath_parser.rb
naitoh: Thanks.
In the following cases, an exception can be detected.
```ruby
> doc = REXML::Document.new("<a><b/></a>")
> REXML::XPath.first(doc, '$typo', nil, {})
lib/rexml/xpath_parser.rb:249:in 'REXML::XPathParser#expr': Undefined variable: $typo (NameError)
```

However, exceptions cannot be detected in the following cases.

```ruby
> REXML::XPath.match(doc, '//nope[@id=$typo]', nil, {})     #=> []
> REXML::XPath.match(doc, '//a[true() or $typo=1]', nil, {}) #=> [<a> ... </>]
```

lib/rexml/xpath_parser.rb
tompng: I added this TODO comment as a TODO comment for future work, not as an unfinished work in this PR.
Do you think it is worth leaving this comment? Is deleting this comment and/or instead file a github issue is better?

lib/rexml/xpath_parser.rb
tompng: Good point. Parse-time validation of all variables in the expression is the right model (and spec-conformant; it shouldn't depend on the document). But that's a bigger change than this PR's contamination fix, and the reference-time NameError I added is really a half-measure toward it.

Do you think the current half-way validation is worth adding for now?
Another idea is to remove this reference-time NameError from this PR, just fallback to `""` for now, and let the whole variable existence check a separate improvement issue. What do you think?

naitoh: > Do you think the current half-way validation is worth adding for now?

I’d prefer to avoid half-baked solutions if possible.

> Another idea is to remove this reference-time NameError from this PR, just fallback to `""` for now, and let the whole variable existence check a separate improvement issue. What do you think?

I agree.
Let’s treat it as a separate improvement issue.

tompng: Applied in cf58ecc1130298b7243755a3a15db4f8fcf52bc2

lib/rexml/xpath_parser.rb
tompng: Created an issue https://github.com/ruby/rexml/issues/350 and removed this TODO comment because
code need to be fixed to handle this issue is not only here, leaving the TODO only on this line may not be a good idea.

lib/rexml/xpath_parser.rb
```diff
@@ -72,7 +72,8 @@ def namespaces=( namespaces={} )
       @namespaces = namespaces
     end
 
-    def variables=( vars={} )
+    def variables=(vars)
+      vars = vars.transform_values { |v| coerce_variable(v) }
       @functions.variables = vars
       @variables = vars
     end
```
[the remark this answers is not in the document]
tompng: Applied in b4e2846432db6e59be9c0cf0ad0da65198f0340e
Although `XPathParser#[]=` is only used from test code (nearly dead code) and `@functions.variables=` has no effect at all (write only, no read). It should be also cleaned up in the future.

naitoh (approved):

lib/rexml/xpath_parser.rb
naitoh: Thank you.
I’ve identified an issue regarding the `REXML::XPath.first` method.

* REXML 3.4.4

```ruby
> doc = REXML::Document.new("<a><b><c/></b><d><e/></d></a>")
> REXML::XPath.match(doc, '$y', nil, { 'x' => 1 }) #=> [nil]
> REXML::XPath.match(doc, '$y') #=> [nil]
> REXML::XPath.first(doc, '$y') #=> nil
> REXML::XPath.first(doc, '//dummy') #=> nil
```

* This PR

```ruby
> REXML::XPath.match(doc, '$y', nil, { 'x' => 1 }) #=> [""]
> REXML::XPath.match(doc, '$y') #=> [""]
> REXML::XPath.first(doc, '$y') #=> ""
> REXML::XPath.first(doc, '//dummy') #=> nil
```

https://docs.ruby-lang.org/ja/latest/class/REXML=3a=3aXPath.html#S_FIRST (Sorry, this comment is in Japanese only.)

```
マッチするノードがない場合には nil を返します。 
```


I think that, as things stand, `REXML::XPath.first(doc, '$y')` returning `nil` because it doesn't match is correct—what do you think?
I think the ideal approach would be to treat this as an `exception`—should we discuss this in a separate issue or PR?

@tompng @kou

tompng: `REXML::XPath.first(doc, '1+2')` doesn't match to a node, and it returns `3`. 
If the point is `XPath.first -> Node | nil`, it's an existing and separate issue.

tompng: I've opened https://github.com/ruby/rexml/issues/353

naitoh: OK.
I see.

naitoh:
Thanks!

PR #344, opened 2026-07-02
Make Ractor-compatible

jhawthorn:
For the most part this just involved freezing Strings or other constants. I tried to aim for the best final state rather than workarounds, so let me know if any changes are desired/anything is too aggressive ❤️. Where possible I changed to `frozen_string_literal: true` instead of simply freezing the required constants.

Two more significant changes are:
* Changed `FunctionsClass` to use `method_defined?(x, false)` instead of maintaining a mutable hash. This required bumping to Ruby 2.6
* Changed `REXML::Security` to use ivars instead of cvars. This module wasn't included anywhere internally so it should work the same. (Ruby 4.1+ will allow cvars to be read from Ractors so this change was only necessary for earlier/current versions, but I think it's faster and better anyways)

### DIFF
--- a/lib/rexml/cdata.rb
+++ b/lib/rexml/cdata.rb
@@ -1,4 +1,4 @@
-# frozen_string_literal: false
+# frozen_string_literal: true
 require_relative "text"
 
 module REXML
--- a/lib/rexml/comment.rb
+++ b/lib/rexml/comment.rb
@@ -1,4 +1,4 @@
-# frozen_string_literal: false
+# frozen_string_literal: true
 require_relative "child"
 
 module REXML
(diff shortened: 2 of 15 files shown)


naitoh:
@jhawthorn
Thanks for this PR.

This PR does not include tests for Ractor, so please add them.

> This required bumping to Ruby 2.6

- This has been addressed in https://github.com/ruby/rexml/pull/347.
  Please rebase.

jhawthorn:
@naitoh Sorry for the delay!

This has been rebased and I've added a couple tests of Ractors. The Ractor tests only run on Ruby 4.0+, as there are known issues on earlier versions (though the tests _should_ pass as far back as 3.1, they might be flaky)

jhawthorn:
This was hanging on windows CI (not sure why, almost certainly an upstream bug we should look into) so I skipped windows with the new tests for now.

kou (approved):
+1

naitoh (approved):
+1

naitoh:
Thanks!

PR #348, opened 2026-07-15
Hide internal classes from documentation

naitoh:
## Why?
`REXML::FunctionsClass` (and the `REXML::Functions` singleton), `REXML::SourceFactory`, `REXML::Source`, `REXML::IOSource`, `REXML::Output` and `REXML::Parsers::XPathParser` are for internal use only, so mark them with `:nodoc:` to exclude them from the generated documentation.

### See
- https://docs.ruby-lang.org/ja/latest/class/REXML=3a=3aFunctions.html
- https://docs.ruby-lang.org/ja/latest/class/REXML=3a=3aSource.html
- https://docs.ruby-lang.org/ja/latest/class/REXML=3a=3aIOSource.html
- https://docs.ruby-lang.org/ja/latest/class/REXML=3a=3aSourceFactory.html
- https://docs.ruby-lang.org/ja/latest/class/REXML=3a=3aOutput.html
- https://docs.ruby-lang.org/ja/latest/class/REXML=3a=3aParsers=3a=3aXPathParser.html

### DIFF
--- a/lib/rexml/functions.rb
+++ b/lib/rexml/functions.rb
@@ -7,7 +7,7 @@ module REXML
   # (2) all method calls from XML will have "-" replaced with "_".
   # Therefore, in XML, "local-name()" is identical (and actually becomes)
   # "local_name()"
-  class FunctionsClass
+  class FunctionsClass # :nodoc:
     @@available_functions = {}
 
     def initialize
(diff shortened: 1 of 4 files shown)


kou:
Thanks.

PR #355, opened 2026-08-14
Don't use Regexp match for `substring-after()` XPath function

voxik:
This is the current behavior:

~~~
$ irb
irb(main):001> require "rexml/document"
=> true
irb(main):002> xml = "<root><x>helloworld</x></root>"
=> "<root><x>helloworld</x></root>"
irb(main):003> doc = REXML::Document.new(xml)
=> <UNDEFINED> ... </>
irb(main):004> 
irb(main):005> xpath1 = "//x[substring-after(text(), 'hello')='world']"
=> "//x[substring-after(text(), 'hello')='world']"
irb(main):006> REXML::XPath.match(doc, xpath1)
=> [<x> ... </>]
irb(main):007> 
irb(main):008> xpath2 = "//x[substring-after(text(), 'hel.o')='world']"
=> "//x[substring-after(text(), 'hel.o')='world']"
irb(main):009> REXML::XPath.match(doc, xpath2)
=> [<x> ... </>]
irb(main):010> REXML::VERSION
=> "3.4.5"
~~~

However, XPath expects `string` match according to specification:

https://www.w3.org/TR/1999/REC-xpath-19991116/#function-substring-after

therefore the `match` for `xpath2` should actually return:

~~~
irb(main):008> xpath2 = "//x[substring-after(text(), 'hel.o')='world']"
=> "//x[substring-after(text(), 'hel.o')='world']"
irb(main):009> REXML::XPath.match(doc, xpath2)
=> []
~~~

Just for comparison, this is similar test case in Python:

~~~
$ python3
Python 3.15.0b4 (3.15.0~b4-1.fc45.x86_64, Jul 18 2026, 00:00:00) [GCC 16.1.1 20260703 (Red Hat 16.1.1-4)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from lxml import etree
... 
... xml = b'<root><x>helloworld</x></root>'
... doc = etree.fromstring(xml)
... 
... xpath1 = "//x[substring-after(text(), 'hello')='world']"
... result1 = doc.xpath(xpath1)
... print('Test 1 (hello):', result1)
... 
... xpath2 = "//x[substring-after(text(), 'hel.o')='world']"
... result2 = doc.xpath(xpath2)
... print('Test 2 (hel.o):', result2)
... 
Test 1 (hello): [<Element x at 0x7fcdbd972b00>]
Test 2 (hel.o): []
~~~

### DIFF
--- a/lib/rexml/functions.rb
+++ b/lib/rexml/functions.rb
@@ -205,8 +205,13 @@ module REXML
     # Kouhei fixed this too
     def substring_after( string, test )
       ruby_string = string(string)
-      return $1 if ruby_string =~ /#{test}(.*)/
-      ""
+      ruby_test = string(test)
+      ruby_index = ruby_string.index(ruby_test)
+      if ruby_index.nil?
+        ""
+      else
+        ruby_string[(ruby_index + ruby_test.length)..-1]
+      end
     end
 
     # Take equal portions of Mike Stok and Sean Russell; mix
--- a/test/functions/test_base.rb
+++ b/test/functions/test_base.rb
@@ -126,6 +126,8 @@ module REXMLTests
     def test_substring_angrez
       testString = REXML::Functions::substring_after("helloworld","hello")
       assert_equal( 'world', testString )
+      testString = REXML::Functions::substring_after("helloworld","hel.o")
+      assert_equal( '', testString )
     end
 
     def test_translate


voxik:
Not sure if the test case is in the right spot, or if it should not actually use the full `REXML::XPath.match`

voxik:
BTW trying to dig into history, the RegExp behavior was introduced by https://github.com/ruby/rexml/commit/0d1b4daa6a4bb97db08303b0d1865b3e0a407dc6 . But that is just some squashed commit. Trying to find the specific reasons for the RegExp, I was able to find the original commit here:

https://web.archive.org/web/20060901105224/http://www.germane-software.com/projects/rexml/changeset/1136

And this is roughly reconstruction of the patch:

~~~diff
From b41dec417658c7804ccbc7db25e929afc81a4f8d Mon Sep 17 00:00:00 2001
From: anonymous
Date: Thu, 15 Aug 2005 05:17:37 +0200
Subject: [PATCH] Fixes roundup issue 43: substring-after bug.

See: http://www.germane-software.com/cgi-bin/roundup/rexml/issue43
---
 lib/rexml/functions.rb | 9 +++------
 test/functions.rb      | 5 +++++
 2 files changed, 8 insertions(+), 6 deletions(-)

diff --git a/lib/rexml/functions.rb b/lib/rexml/functions.rb
index 2550cb3..33895ca 100644
--- a/lib/rexml/functions.rb
+++ b/lib/rexml/functions.rb
@@ -158,12 +158,9 @@ module REXML
     # Kouhei fixed this too
     def Functions::substring_after( string, test )
       ruby_string = string(string)
-      ruby_index = ruby_string.index(string(test))
-      if ruby_index.nil?
-        ""
-      else
-        ruby_string[ ruby_index+1..-1 ]
-      end
+      test_string = string(test)
+      return $1 if ruby_string =~ /#{test}(.*)/
+      ""
     end
 
     # Take equal portions of Mike Stok and Sean Russell; mix
diff --git a/test/functions.rb b/test/functions.rb
index 16dec57..82e08a6 100644
--- a/test/functions.rb
+++ b/test/functions.rb
@@ -100,4 +100,9 @@
 end
 
+def test_substring_angrez
+   testString = REXML::Functions::substring_after("helloworld","hello")
+   assert_equal( 'world', testString )
+end
+
 def test_translate 
    source = <<-EOF
~~~

But it seems the original Roundup tracker with the details is lost :(

voxik:
I also think that this implementation might be faster then the RegExp, but I have not tried and it was not the reason for this PR

lib/rexml/functions.rb
```diff
@@ -205,8 +205,10 @@ def substring_before( string, test )
     # Kouhei fixed this too
     def substring_after( string, test )
       ruby_string = string(string)
-      return $1 if ruby_string =~ /#{test}(.*)/
-      ""
+      needle = string(test)
+      index = ruby_string.index(needle)
+      return "" if index.nil?
+      ruby_string[(index + needle.length)..-1]
```
kou: Could you use the same style?


```suggestion
      ruby_test = string(test)
      ruby_index = ruby_string.index(ruby_test)
      if ruby_index.nil?
        ""
      else
        ruby_string[(ruby_index + ruby_test.length)..-1]
      end
```

voxik: Sure. Done

kou:
Thanks.

PR #356, opened 2026-08-14
Fix document order in REXML::XPathParser.sort

naitoh:
`REXML::XPathParser.sort` keys each node on its index under each ancestor, but two different nodes could end up with the same key, and ties are then broken arbitrarily by an unstable sort.

The walk stopped at the root element, so the root and everything outside it -- comments and PIs at document level -- all keyed on an empty array:

```
    <!--c1--><?pi1?><root/><!--c2--><?pi2?>
```

    //comment() | //processing-instruction()
      was: c1, c2, pi1, pi2   want: c1, pi1, c2, pi2

Walk up to the document instead, so those nodes are ordered against each other.

An attribute borrowed the key of the element carrying it, so the element and its attributes tied:

```
    <root><a x="1" y="2"><b/></a></root>
```

    //a/@* | //a | //a/*
      was: @x, @y, <a>, <b>   want: <a>, @x, @y, <b>

Extend an attribute's key past its element's with ATTRIBUTE_POSITION.  A child index is never negative, so -1 lands the attributes after `<a>` and ahead of `<b>`, which is where document order wants them.

The attributes of one element tied with each other too, which went unnoticed because a small enough sort leaves its input alone:

    <a z="1" m="2" b="3"/>       //a/@*  ->  z, m, b
    <a z="1" ... 26 attributes/> //a/@*  ->  scrambled

XPath 1.0 leaves their relative order implementation dependent, but document order is a total ordering, so it has to be decided.  Key them on where they were written, which is the order the small case already appeared to have.  Finding that out means walking the attribute list, so index the whole list at once and remember it for the rest of the sort; asking per attribute would make sorting quadratic in the number of attributes an element carries, which is something a document gets to choose.

The ancestor walk becomes a method of its own, since both branches of the key need it.  It, attribute_position and ATTRIBUTE_POSITION are private rather than public names marked :nodoc:, sort being the only part callers outside the class use.

### DIFF
--- a/lib/rexml/xpath_parser.rb
+++ b/lib/rexml/xpath_parser.rb
@@ -804,6 +804,11 @@ module REXML
       trace(:leave, tag, *args)
     end
 
+    # Sorts before any real child index, so that the attributes of an element
+    # come after the element itself but before its children.
+    ATTRIBUTE_POSITION = -1
+    private_constant :ATTRIBUTE_POSITION
+
     # Reorders an array of nodes so that they are in document order
     # It tries to do this efficiently.
     #
@@ -815,23 +820,52 @@ module REXML
     def self.sort(array_of_nodes)
       return array_of_nodes if array_of_nodes.size <= 1
 
-      new_arry = []
-      array_of_nodes.each { |node|
-        node_idx = []
-        np = node.node_type == :attribute ? node.element : node
-        while np.parent and np.parent.node_type == :element
-          node_idx << np.parent.index( np )
-          np = np.parent
+      attribute_positions = {}.compare_by_identity
+      array_of_nodes.sort_by do |node|
+        if node.node_type == :attribute
+          # An attribute has no place of its own in the child tree, so its key
+          # extends that of the element carrying it.
+          ancestor_indexes(node.element) <<
+            ATTRIBUTE_POSITION << attribute_position(node, attribute_positions)
+        else
+          ancestor_indexes(node)
         end
-        new_arry << [ node_idx.reverse, node ]
-      }
-      ordered = new_arry.sort_by do |index, node|
-        index
-      end
-      ordered.collect do |_index, node|
-        node
       end
     end
+
+    # The index the node holds under each of its ancestors, outermost first.
+    def self.ancestor_indexes(node)
+      indexes = []
+      # Walk all the way up to the document.  Stopping at the root element
+      # would leave every node outside it, and the root itself, with the same
+      # empty key, and ties are then broken arbitrarily.
+      while (parent = node.parent)
+        indexes << parent.index(node)
+        node = parent
+      end
+      indexes.reverse!
+    end
+    private_class_method :ancestor_indexes
+
+    # Where the attribute sits among the attributes of its element.  XPath 1.0
+    # leaves the relative order of those implementation dependent, but document
+    # order is a total ordering, so they do need one; this keeps them in the
+    # order they were written in.
+    def self.attribute_position(attribute, positions)
+      position = positions[attribute]
+      return position if position
+
+      # Index the whole attribute list at once.  A node set often holds every
+      # attribute of an element, and looking each one up on its own would make
+      # sorting quadratic in the number of attributes.
+      i = 0
+      attribute.element.attributes.each_attribute do |other|
+        positions[other] ||= i
+        i += 1
+      end
+      positions[attribute]
+    end
+    private_class_method :attribute_position
 
     # Scanner for descendant-or-self axis
     def descendant_or_self(nodeset, tester, selector)
--- a/test/xpath/test_base.rb
+++ b/test/xpath/test_base.rb
@@ -1,8 +1,11 @@
 # frozen_string_literal: false
 
+require "core_assertions"
+
 module REXMLTests
   class TestXPathBase < Test::Unit::TestCase
     include Helper::Fixture
+    include Test::Unit::CoreAssertions
     include REXML
     SOURCE = <<-EOF
       <a id='1'>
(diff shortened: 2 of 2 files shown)


lib/rexml/xpath_parser.rb
```diff
@@ -815,23 +820,51 @@ def leave(tag, *args)
...
+      # attribute of an element, and looking each one up on its own would make
+      # sorting quadratic in the number of attributes.
+      i = 0
+      attribute.element.attributes.each_attribute do |other|
+        positions[other] ||= i
+        i += 1
       end
+      positions[attribute]
     end
+    private_class_method :ancestor_indexes, :attribute_position
```
kou: Could you move `:ancestor_indexes` just after `def self.ancestor_indexes`?

Or could you use `private_class_method def self.ancestor_indexes`?

test/xpath/test_base.rb
```diff
@@ -1564,6 +1564,85 @@ def test_reverse_axis_function_argument_sort
...
     end
 
+    def test_document_order_top_level_nodes
+      # Nodes outside the root element are still ordered against each other.
+      doc = Document.new("<!--c1--><?pi1?><root/><!--c2--><?pi2?>")
+      nodes = XPath.match(doc, "//comment() | //processing-instruction()")
+      assert_equal(["c1", "pi1", "c2", "pi2"],
+                   nodes.collect {|node|
+                     node.node_type == :comment ? node.string : node.target
+                   })
```
kou: Could you extract code that stringifies nodes?

```ruby
def stringify_nodes(nodes)
  node.collect do |node|
    ...
  end
end
```

test/xpath/test_base.rb
```diff
@@ -1564,6 +1564,85 @@ def test_reverse_axis_function_argument_sort
...
+    def test_document_order_element_precedes_its_attributes
+      doc = Document.new("<root><a x='1' y='2'/></root>")
+      # The relative order of attributes of one element is implementation
+      # dependent, but they must all come after the element that carries them,
+      # whichever way round the union is written.
+      ["//a | //a/@*", "//a/@* | //a"].each do |path|
+        nodes = XPath.match(doc, path)
+        assert_equal("a", nodes.first.name, path)
+        assert_equal(["x", "y"], nodes[1..-1].collect(&:name).sort, path)
+      end
```
kou: Could you use separated tests instead of `[...].each` for easy to debug?

test/xpath/test_base.rb
```diff
@@ -1564,6 +1564,85 @@ def test_reverse_axis_function_argument_sort
...
+
+    def test_document_order_attributes_of_one_element
+      # XPath 1.0 leaves the relative order of the attributes of one element
+      # implementation dependent, but document order is a total ordering, so it
+      # still has to be decided: they come out in the order they were written.
+      names = ("a".."z").to_a.reverse
+      attributes = names.collect {|name| "#{name}='1'" }.join(" ")
+      doc = Document.new("<root><a #{attributes}/></root>")
+      assert_equal(names, XPath.match(doc, "//a/@*").collect(&:name))
+      assert_equal("1", XPath.match(doc, "string(//a/@*)").first)
```
kou: Why do we need this?

test/xpath/test_base.rb
```diff
@@ -1564,6 +1564,85 @@ def test_reverse_axis_function_argument_sort
...
+    end
+
+    def test_document_order_many_attributes_of_one_element
+      # Ordering the attributes of an element must not cost more than the
+      # attributes themselves: one whole list is indexed per element, not one
+      # list per attribute.
+      count = 2000
```
kou: Do we need 2000 attributes here?

test/xpath/test_base.rb
```diff
@@ -1564,6 +1564,85 @@ def test_reverse_axis_function_argument_sort
...
+      attributes = (0...count).collect {|i| "a#{i}='1'" }.join(" ")
```
kou: `count.times.collect` is easier to read than `(0...count).collect`.

test/xpath/test_base.rb
tompng: `XPath.match` in this test takse 0.005s in my machine.
If we simplify `attribute_position` to the code below, `XPath.match` still works in 0.25s
```ruby
def self.attribute_position(attribute, _positions)
  attribute.element.attributes.each_attribute.find_index(attribute)
end
```
So we need to either increase this number, add a timeout (much smaller than 0.25s), or use `assert_linear_performance` to detect quadratic time regression.

lib/rexml/xpath_parser.rb
naitoh: OK, I see.

test/xpath/test_base.rb
naitoh: OK, I see.

test/xpath/test_base.rb
naitoh: OK, I see.

test/xpath/test_base.rb
naitoh: Thanks.
I will use `assert_linear_performance` to modify the code so that it can detect the performance issue I was trying to identify.

test/xpath/test_base.rb
naitoh: OK, I see.

test/xpath/test_base.rb
naitoh: Sorry.
The test `assert_equal(“1”, XPath.match(doc, “string(//a/@*)”).first)` isn't working as intended, so I'll fix it.

naitoh:
I have addressed the issues pointed out by PR
Please review it again.

test/xpath/test_base.rb
```diff
@@ -1564,6 +1567,89 @@ def test_reverse_axis_function_argument_sort
...
+      doc = Document.new(source)
+      nodes = XPath.match(doc, "//a/node()")
+      assert_equal(["t0", "b#0", "u0", "t1", "b#1", "u1"],
+                   nodes.collect {|node|
+                     if node.node_type == :element
+                       "b##{node.attributes['id']}"
+                     else
+                       node.value
+                     end
+                   })
```
kou: Can we use `stringify_nodes` here too?

naitoh: OK, I see.

naitoh: I've made the correction.
Please review it again.

test/xpath/test_base.rb
```diff
@@ -1564,6 +1567,82 @@ def test_reverse_axis_function_argument_sort
...
+      # Attributes of different elements are ordered by their owning elements.
+      doc = Document.new("<root><a id='1'/><a id='2'/><a id='3'/></root>")
+      assert_equal(["1", "2", "3"], XPath.match(doc, "//a/@id").collect(&:value))
+    end
+
+    def test_document_order_mixed_text_and_element_children
+      source = <<-XML
+<root>
+  <a>t0<b/>u0</a>
+  <a>t1<c/>u1</a>
```
kou: BTW, what does "u" mean? Alphabet order? (t -> u?)

"t0" -> "before0"
"u0" -> "after0"
may be easy to understand.

naitoh: Thanks for pointing that out.

Since the “u” had no meaning, I corrected it as you pointed out.

kou:
Thanks.

PR #357, opened 2026-08-15
Fix Element#attribute when a namespace URI has multiple bindings

### LINKED ISSUE #346: `Element#attribute` fails to find attribute when namespace URL is duplicated
```ruby
xml = "<root xmlns='url1' xmlns:ns1='url1' xmlns:ns2='url2' xmlns:ns3='url2' a='' ns1:a='' ns1:b='' ns2:c='' ns3:d=''/>"
node = REXML::Document.new(xml).root

node.attribute 'a', 'url1'
# => a='' (ns1:a should match)
node.attribute 'b', 'url1'
# => nil (ns1:b='' should match)
node.attribute 'c', 'url2'
# => ns2:c='' (OK)
node.attribute 'd', 'url2'
# => nil (ns3:d='' should match)

# Ref:
node.attributes.get_attribute_ns '', 'a'     #=> a=''
node.attributes.get_attribute_ns 'url1', 'a' #=> ns1:a=''
node.attributes.get_attribute_ns 'url1', 'b' #=> ns1:b=''
node.attributes.get_attribute_ns 'url2', 'c' #=> ns2:c=''
node.attributes.get_attribute_ns 'url2', 'd' #=> ns3:d=''
```

Result of `node.attribute 'b', 'url1'` has changed in #335 as a bugfix (DOM spec violation) of `Attributes#get_namespace`.
`node.attribute 'a', 'url1'` case is related to #151.

The easiest way to fix this is to simply use:
```ruby
def attribute( name, namespace=nil )
  if namespace
    attributes.get_attribute_ns(namespace, name)
  else
    attributes.get_attribute(name)
  end
end
```
However, it may introduce incompatibility mentioned in #151.
If this method is fixed as above, I think we don't need to deprecate this method (discussed in #151) because it will be a useful method: a superset of `attributs.get_attribute_ns` and `attributs.get_attribute`.

sjh9714:
Fixes #346.

## Motivation

`Element#attribute` can miss a namespaced attribute when the same namespace URI is bound to multiple prefixes. Lookups with prefixed attributes could also depend on declaration order.

`Attributes#get_attribute_ns` resolves the namespaced match consistently. This change makes `Element#attribute` use that result first.

## Changes

- Keep the direct `get_attribute` lookup when no namespace is supplied.
- When a namespace is supplied, call `get_attribute_ns(namespace, name)` first.
- If that finds nothing, preserve the unprefixed fallback when the requested URI is the default namespace or is not declared on the element.

A matching result from `get_attribute_ns` now wins when both prefixed and unprefixed attributes exist.

## Tests

- Add coverage for a namespace URI bound to multiple prefixes.
- Add coverage for all four relevant namespace and attribute declaration orders.

```text
ruby test/run.rb
```

799 tests, 0 failures.

### DIFF
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -1269,22 +1269,17 @@ module REXML
     #   document.root.attribute("x", "a") # => a:x='a:x'
     #
     def attribute( name, namespace=nil )
-      prefixes = namespace ? namespaces.select {|_, uri| uri == namespace}.keys : []
+      return attributes.get_attribute( name ) if namespace.nil?
 
-      # An unprefixed attribute is used for the default namespace.
-      if prefixes.empty? or prefixes.include?( 'xmlns' )
-        ret_val = attributes.get_attribute( name )
-        return ret_val unless ret_val.nil?
+      ret_val = attributes.get_attribute_ns( namespace, name )
+      return ret_val unless ret_val.nil?
+
+      # Kept for compatibility: an unprefixed attribute also matches when the
+      # requested URI is the default namespace or is not declared on this
+      # element, even though such an attribute is not in that namespace.
+      if namespace == namespaces[ 'xmlns' ] or !namespaces.has_value?( namespace )
+        attributes.get_attribute( name )
       end
-
-      # The same namespace URI may be bound to multiple prefixes.
-      prefixes.each do |prefix|
-        next if prefix == 'xmlns'
-        ret_val = attributes.get_attribute( "#{prefix}:#{name}" )
-        return ret_val unless ret_val.nil?
-      end
-
-      nil
     end
 
     # :call-seq:
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -1269,20 +1269,22 @@ module REXML
     #   document.root.attribute("x", "a") # => a:x='a:x'
     #
     def attribute( name, namespace=nil )
-      prefix = namespaces.key(namespace) if namespace
-      prefix = nil if prefix == 'xmlns'
+      prefixes = namespace ? namespaces.select {|_, uri| uri == namespace}.keys : []
 
-      ret_val =
-        attributes.get_attribute( prefix ? "#{prefix}:#{name}" : name )
+      # An unprefixed attribute is used for the default namespace.
+      if prefixes.empty? or prefixes.include?( 'xmlns' )
+        ret_val = attributes.get_attribute( name )
+        return ret_val unless ret_val.nil?
+      end
 
-      return ret_val unless ret_val.nil?
-      return nil if prefix.nil?
+      # The same namespace URI may be bound to multiple prefixes.
+      prefixes.each do |prefix|
+        next if prefix == 'xmlns'
+        ret_val = attributes.get_attribute( "#{prefix}:#{name}" )
+        return ret_val unless ret_val.nil?
+      end
 
-      # now check that prefix'es namespace is not the same as the
-      # default namespace
-      return nil unless ( namespaces[ prefix ] == namespaces[ 'xmlns' ] )
-
-      attributes.get_attribute( name )
+      nil
     end
 
     # :call-seq:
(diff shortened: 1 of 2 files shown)


lib/rexml/element.rb
```diff
@@ -1269,20 +1269,22 @@ def [](name_or_index)
     #   document.root.attribute("x", "a") # => a:x='a:x'
     #
     def attribute( name, namespace=nil )
```
tompng, line 1271: For this XML:
```ruby
root1 = REXML::Document.new("<root xmlns:ns1='url1' xmlns='url1' a='A' ns1:a='NS1A'/>").root
root2 = REXML::Document.new("<root a='A' xmlns:ns1='url1' xmlns='url1' ns1:a='NS1A'/>").root
root3 = REXML::Document.new("<root xmlns:ns1='url1' xmlns='url1' ns1:a='NS1A' a='A'/>").root
root4 = REXML::Document.new("<root xmlns='url1' xmlns:ns1='url1' ns1:a='NS1A' a='A'/>").root
```

`root.attributes.get_attribute_ns('url1', 'a').expanded_name` always returns `"ns1:a"`.
But `attribute('a', 'url1')` is different.
```ruby
[root1, root2, root3, root4].map do |root|
  root.attribute('a', 'url1').expanded_name
end
# Master: ["ns1:a", "ns1:a", "ns1:a", "a"]
#   (the result depends on the declaration order of xmlns / xmlns:ns1 — this is a bug)
# This PR: ["a", "a", "a", "a"]
```
The behavior is changed, and it is changed to the opposite way from `get_attribute_ns`. Note that the PR description says "No lookup that previously returned an attribute returns a different one", but root1–root3 are counterexamples. It's worth describing this change in the PR description.
Though the expected behavior should be decided by the maintainer (@naitoh), I think it may be nice to align with `get_attribute_ns`.

If we're going to mainly align with `get_attribute_ns`, how about first calling `get_attribute_ns`, and falling back to a custom logic if nothing is found? The compatibility-oriented loose behaviors (matching an unprefixed attribute for the default namespace or for an undeclared namespace URI) can live in the fallback, and it would make it much easier to see how this method differs from `get_attribute_ns`.

sjh9714:
Restructured in ac3e8b8 as you suggested. `attribute` now calls `attributes.get_attribute_ns` first, and the compatibility-oriented behaviors live in the fallback, so an unprefixed attribute still matches when the requested URI is the default namespace or is not declared on the element.

Your four examples now all return `ns1:a`, matching `get_attribute_ns` regardless of declaration order, and I added a regression test for exactly those documents. The existing test's `attribute("a", "url1")` expectation changed from `a` to `ns1:a` accordingly.

On the description's compatibility claim, you're right, root1-root3 were counterexamples. With this revision the remaining difference from master is narrower. A lookup can change only when the requested URI is the default namespace and both an unprefixed and a prefixed attribute are present. Master returned one or the other depending on xmlns declaration order, and this branch now always returns the prefixed one, like `get_attribute_ns`. The loose unprefixed fallbacks behave as before when no prefixed attribute matches. I can flip that priority if @naitoh prefers unprefixed-first for the default namespace.

Ran `ruby test/run.rb`, 799 tests, 0 failures.

lib/rexml/element.rb
naitoh: @kou 
- Since #151 caused incompatibility with existing test code, we prioritized compatibility by only adding documentation in #360; however, ac3e8b8aca9e52ad3a2ac09f49ca4754b2736666 does not affect existing test code.

- Also, `root1`, `root2`, `root3`, and `root4` should be fixed so that they return the same result. The response for `ns1:a` in ac3e8b8aca9e52ad3a2ac09f49ca4754b2736666 (which is the same as get_attribute_ns) seems reasonable—what do you think?

kou: OK. Let's do it.

kou (approved):
+1

naitoh (approved):

naitoh:
@sjh9714
We think ac3e8b8aca9e52ad3a2ac09f49ca4754b2736666 is good.
Could you please update the PR description to match ac3e8b8aca9e52ad3a2ac09f49ca4754b2736666?

sjh9714:
Updated the PR description to match `ac3e8b8` and removed the stale behavior claims. It now describes the `get_attribute_ns` lookup, the compatibility fallback and the reported 799-test result. Thanks.

naitoh:
Thanks!

PR #360, opened 2026-08-17
Document how Element#attribute matches a namespace

naitoh:
GitHub: Fix GH-151

`REXML::Element#attribute` matches its +namespace+ argument loosely: an unprefixed attribute is taken to be in the default namespace, and a namespace that no prefix is bound to falls back to the unprefixed attribute.

    <root xmlns='ns0' a='a'/>

    attribute("a", "ns0")    -> a='a'
    attribute("a", "nosuch") -> a='a'

The XML Namespaces specification says an unprefixed attribute has no namespace and does not take the default one, so neither of those should match.  The
behavior is kept for compatibility -- it is what tickets 102 and 121 asked for, and the tests for both are still in place -- but nothing said so, which has left people reading the method unsure whether what they saw was the contract or a bug.

Say it in the documentation of both methods, and point each at the other: REXML::Attributes#get_attribute_ns matches as the XML Namespaces specification says and is what to reach for when the namespace has to be matched strictly.  It has behaved that way at least as far back as GH-151, but that discussion never mentioned it.

Reported by Hiroya Fujinami. Thanks!!!

### DIFF
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -1268,6 +1268,22 @@ module REXML
     #   document.root.attribute("x")      # => x='x'
     #   document.root.attribute("x", "a") # => a:x='a:x'
     #
+    # This method matches +namespace+ loosely.  An unprefixed attribute is
+    # taken to be in the default namespace, and a +namespace+ that no prefix
+    # is bound to falls back to the unprefixed attribute:
+    #
+    #   xml_string = "<root xmlns='ns0' a='a'/>"
+    #   document = REXML::Document.new(xml_string)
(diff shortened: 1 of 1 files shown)


naitoh:
Based on the discussion below, we will update the documentation for `REXML::Element#attribute` and `REXML::Attributes#get_attribute_ns` to include the current behavior regarding namespace specification as part of the specification.

- https://github.com/ruby/rexml/pull/151#issuecomment-2177803175
- https://github.com/ruby/rexml/issues/346
- https://github.com/ruby/rexml/pull/357

PR #363, opened 2026-08-25
Deprecate unused `REXML::Text.read_with_substitution`

kou:
We don't want to maintain unused code.

Use `REXML::Text.unnormalize` instead.


naitoh (approved):

naitoh:
Thanks!

PR #366, opened 2026-08-30
Reject underscored XPath function names

### LINKED ISSUE #352: XPath underscored function name is wrongly accepted
```ruby
# local-name: OK
REXML::XPath.match REXML::Document.new('<foo/>'), 'local-name(*)'
#=> ["foo"]

# local_name: Should be rejected
REXML::XPath.match REXML::Document.new('<foo/>'), 'local_name(*)'
#=> ["foo"]
```

sjh9714:
## Summary

Fixes #352.

Reject XPath function names containing underscores instead of treating them as
aliases for hyphenated standard functions. Standard names such as
`local-name()` continue to parse normally.

## Changes

- Stop underscored names from being parsed as XPath function calls.
- Update XPath fixtures to use standard hyphenated names and add regression
  coverage.

## Testing

- `ruby -Itest -Ilib test/parser/test_xpath.rb --name=test_function_with_underscore`
- `rake test`

### DIFF
--- a/lib/rexml/parsers/xpathparser.rb
+++ b/lib/rexml/parsers/xpathparser.rb
@@ -13,6 +13,10 @@ module REXML
       include XMLTokens
       LITERAL    = /^'([^']*)'|^"([^"]*)"/u
 
+      def initialize(strict: false)
+        @strict = strict
+      end
+
       def namespaces=( namespaces )
         Functions::namespace_context = namespaces
         @namespaces = namespaces
@@ -653,7 +657,7 @@ module REXML
           #arry << @variables[ varname ]
         when /^(\w[-\w]*)(?:\()/
           fname = $1
-          return path if fname.include?("_")
+          return path if @strict && fname.include?("_")
           tmp = $'
           return path if fname =~ NT
           path = tmp
--- a/lib/rexml/parsers/xpathparser.rb
+++ b/lib/rexml/parsers/xpathparser.rb
@@ -653,6 +653,7 @@ module REXML
           #arry << @variables[ varname ]
         when /^(\w[-\w]*)(?:\()/
           fname = $1
+          return path if fname.include?("_")
           tmp = $'
           return path if fname =~ NT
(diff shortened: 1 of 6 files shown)


kou:
I want to keep backward compatibility as much as possible.

How about rejecting `_` only when a user specify an option explicitly?

sjh9714:
Updated this so underscored function names remain accepted by default and are rejected only when the existing `strict: true` option is passed. The low-level XPath parser now receives that option, and the tests cover both default compatibility and strict rejection.

Ran:
- `ruby -Ilib:test test/parser/test_xpath.rb`
- `ruby -Ilib:test test/functions/test_base.rb`
- `ruby -Ilib:test test/run.rb`

lib/rexml/parsers/xpathparser.rb
```diff
@@ -653,6 +657,7 @@ def PrimaryExpr path, parsed
           #arry << @variables[ varname ]
         when /^(\w[-\w]*)(?:\()/
           fname = $1
+          return path if @strict && fname.include?("_")
```
tompng, line 660: The mechanism that this `return path` rejects the xpath, is complicated. (See LocationPath → RelativeLocationPath → NodeTest called when `'local_name()'` is passed)

I think underscored function names are invalid not because it has underscore, but because it's not registered as a function in XPath 1.0 spec.
Both examples in the code below is in the same situation: using unknown function, so the result (return value or raised error) should be the same.
```ruby
REXML::XPath.match(REXML::Document.new('<root/>'), 'local_name()')
REXML::XPath.match(REXML::Document.new('<root/>'), 'localname()')
```

From a viewpoint of module's responsibility, I personally think this check shouldn't be done in REXML::Parsers:XPathParser, but in one of:
- `REXML::XPathParser`: layer that converts `'-'` to `'_'`, function name to a method name
- `REXML::FunctionsClass`: layer that knows all available functions, checks if a method name is registered as a function

PR #367, opened 2026-08-31
Follow up GH-357 in the documentation of Element#attribute

naitoh:
GH-360 documented `REXML::Element#attribute` as matching a namespace loosely, and said an unprefixed attribute is taken to be in the default namespace.  GH-357 made the method ask
REXML::Attributes#get_attribute_ns first, so that is now only true when nothing matches strictly:

    <root xmlns='ns0' xmlns:p='ns0' a='A' p:a='PA'/>

    attribute("a", "ns0")   ->  p:a='PA', not a='A'

The two examples in the note still hold, because neither document has a prefixed attribute of the same local name for the strict match to find. It is the rule stated around them that no longer holds: an unprefixed attribute is taken to be in the default namespace only when nothing matches strictly.

Describe the method the way it now works -- strict first, then the compatibility fallback -- and add the example above, which is the only place the change is visible.  Say of get_attribute_ns that it matches strictly and only strictly, rather than pointing at it as the way to match strictly, now that Element#attribute starts there too.

### DIFF
--- a/lib/rexml/element.rb
+++ b/lib/rexml/element.rb
@@ -1268,9 +1268,11 @@ module REXML
     #   document.root.attribute("x")      # => x='x'
     #   document.root.attribute("x", "a") # => a:x='a:x'
     #
-    # This method matches +namespace+ loosely.  An unprefixed attribute is
-    # taken to be in the default namespace, and a +namespace+ that no prefix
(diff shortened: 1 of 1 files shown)


kou:
Thanks.

PR #362, opened 2026-08-23
Remove unused QuickPath

kou:
GitHub: Fix GH-250

It's not used, tested and documented.

PR #368, opened 2026-09-01
Cleanup unused accessor methods inf REXML::Functions

tompng:
Removes unused `Functions.variables`. It was write only, never read.
Removes `Functions.namespace_context`. It was only used from QuickPath, and QuickPath has been removed in #362


kou:
Good catch.

PR #372, opened 2026-09-09
Reject invalid character references in entity expansion

kou:
Text.expand and BaseParser#unnormalize packed numeric character references directly with Array#pack without validating the code point. Unlike literal references, which Text.check validates, the entity-expansion path (Text#value, attribute values, and entity values) could therefore:

  * emit XML-forbidden characters such as NUL, other control characters, and the noncharacters U+FFFE / U+FFFF, and
  * emit invalid UTF-8, or leak a bare RangeError from Array#pack, for code points beyond U+10FFFF.

These happen lazily at value-access time, so the document parses and only fails (or silently corrupts) later.

Validate the code point via the new Text.expand_character_reference against VALID_CHAR and raise REXML::ParseException for invalid values, so the expansion path behaves consistently with the literal path. The error message echoes the original reference verbatim to keep decimal and hexadecimal forms distinguishable.

### DIFF
--- a/lib/rexml/text.rb
+++ b/lib/rexml/text.rb
@@ -387,12 +387,26 @@ module REXML
       }
     end
 
+    # Expands a numeric character reference's code point to a String.
+    # Code points that are not valid XML characters are rejected, matching
+    # the validation Text.check applies to literal references. This keeps the
+    # entity-expansion path from emitting XML-forbidden characters (NUL,
+    # control chars, U+FFFE/U+FFFF, code points beyond U+10FFFF, ...).
+    def Text.expand_character_reference(code_point, reference)
+      case code_point
+      when *VALID_CHAR
+        [code_point].pack('U*')
+      else
+        raise ParseException.new("Illegal character reference: <#{reference}>")
+      end
+    end
+
     def Text.expand(ref, doctype, filter, expanding: nil)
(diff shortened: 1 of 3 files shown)


lib/rexml/text.rb
```diff
@@ -387,12 +387,26 @@ def Text::unnormalize( string, doctype=nil, filter=nil, illegal=nil, entity_expa
...
+    # Code points that are not valid XML characters are rejected, matching
+    # the validation Text.check applies to literal references. This keeps the
+    # entity-expansion path from emitting XML-forbidden characters (NUL,
+    # control chars, U+FFFE/U+FFFF, code points beyond U+10FFFF, ...).
+    def Text.expand_character_reference(code_point, reference)
+      case code_point
+      when *VALID_CHAR
+        [code_point].pack('U*')
+      else
+        raise ParseException.new("Illegal character reference: <#{reference}>")
```
kou, line 400: Is `ParseException` OK...?

naitoh: ParseException looks right to me. :)

naitoh (approved):
+1

lib/rexml/text.rb
kou: OK. Let's keep this.

